PR #6, opened 2024-08-27
Publisher Agent

harryrackmil:
* Hosts a websocket server for receiving prices from a publisher
* Also supports specifying a websocket to pull price data from
* Configure whether you want to sign every message or only based on clock/delta conditions
* EVM signatures are entirely in Go, the slow parts of Stark signing are shipped out to a compiled Rust function


cli/stork-publisher-agent/util_test.go
```diff
@@ -0,0 +1,21 @@
...
+	quantizedPrice := FloatToQuantizedPrice(72147.681412670819)
+	expectedQuantizedPrice := QuantizedPrice("72147681412670819000000")
+	assert.Equal(t, expectedQuantizedPrice, quantizedPrice)
+
+	quantizedPrice = FloatToQuantizedPrice(3.33595034988)
+	expectedQuantizedPrice = "3335950349880000000"
+	assert.Equal(t, expectedQuantizedPrice, quantizedPrice)
+
+	// todo: this case fails
+	//quantizedPrice = FloatToQuantizedPrice(3.335950349883)
```
harryrackmil, line 18: the conversion from float to quantized price introduces some rounding error - this same issue happens in the Rust code.

Might make more sense to take in a string representation of price or allow users to submit already-quantized prices to the agent. Leaving as is for now because price service and Hopper both output floats.

cli/stork-publisher-agent.go
```diff
@@ -0,0 +1,192 @@
...
+const OracleIdFlag = "oracle-id"
+const PrivateKeyFlag = "private-key"
+const PublicKeyFlag = "public-key"
+
+// optional
+const ClockUpdatePeriodFlag = "clock-update-period"
+const DeltaUpdatePeriodFlag = "delta-update-period"
+const ChangeThresholdPercentFlag = "change-threshold-percent"
+const IncomingWsPortFlag = "incoming-ws-port"
+const OutgoingWsUrlFlag = "outgoing-ws-url"
```
harryrackmil, line 34: this will be replaced with a call to the Stork Registry in the future

cli/stork-publisher-agent/signer.go
```diff
@@ -0,0 +1,294 @@
...
+}
+
+func createBufferFromBigInt(i *big.Int) *C.uint8_t {
+	bytes := make([]byte, 32)
+	i.FillBytes(bytes)
+	return (*C.uint8_t)(unsafe.Pointer(&bytes[0]))
+}
+
+func (s *Signer[T]) SignStark(assetHexPadded string, quantizedPrice QuantizedPrice, timestampNs int64) (*TimestampedSignature[T], error) {
+	assetInt, _ := new(big.Int).SetString(strip0x(assetHexPadded), 16)
```
harryrackmil, line 243: move to rust

cli/stork-publisher-agent/log.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/rs/zerolog/log"
+	"net/http"
+)
+
+func BaseAppLogger() zerolog.Logger {
+
+	return log.With().Str("application", "stork-publisher-agent").
+		Logger().
+		Hook(zerolog.HookFunc(func(e *zerolog.Event, l zerolog.Level, msg string) {
+			// for capture by cloudwatch log filter, should be in line with python logger levels
```
akawalsky, line 14: This is a little internal Stork specific, doesn't necessarily need to be this opinionated for external users

cli/stork-publisher-agent/runner.go
```diff
@@ -0,0 +1,281 @@
...
+		}
+
+		_, messageBytes, err := incomingWsConn.ReadMessage()
+		if err != nil {
+			r.logger.Error().Err(err).Msgf("Failed to read connection message from pull-based WebSocket: %v", err)
+		}
+		r.logger.Info().Msgf("Received connection message: %s", messageBytes)
+
+		if len(subscriptionRequest) > 0 {
+			r.logger.Info().Msgf("Sending subscription request: %s", subscriptionRequest)
```
akawalsky, line 158: Lets do a quick audit of all of these actually

cli/stork-publisher-agent.go
```diff
@@ -0,0 +1,204 @@
...
+var publisherAgentCmd = &cobra.Command{
+	Use:   "publisher-agent",
+	Short: "Start a process to sign price updates and make them available to Stork",
+	RunE:  runPublisherAgent,
+}
+
+// required
+const SignatureTypeFlag = "signature-type"
+const OracleIdFlag = "oracle-id"
+const PrivateKeyFlag = "private-key"
```
akawalsky, line 26: Stark and Evm keys are differently structured pretty sure. Flagging that this is an issue, lets discuss offline.

cli/stork-publisher-agent/util_test.go
akawalsky: Do we need the data to be in terms of float at any point? Maybe we should instead convert everything to "Big" by default

akawalsky (commented):
Some initial comments, looking good so far

cli/stork-publisher-agent/model.go
```diff
@@ -0,0 +1,102 @@
...
+	SignatureType  string
+	QuantizedPrice string
+)
+
+const EvmSignatureType = SignatureType("evm")
+const StarkSignatureType = SignatureType("stark")
+
+// Incoming
+
+type PriceUpdate struct {
```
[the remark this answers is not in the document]
harryrackmil: yea I like `Value` over `Price`

cli/stork-publisher-agent/runner.go
harryrackmil: went through all the log statements and everything expected to Debug, so Info will only be expected failures like websockets closing or being intentionally removed and real problems will be warn or error.

cli/stork-publisher-agent/runner.go
```diff
@@ -0,0 +1,281 @@
...
+		connId := ConnectionId(uuid.New().String())
+
+		r.logger.Info().Str("auth_token", string(authToken)).Str("conn_id", string(connId)).Msg("adding subscriber websocket")
+
+		websocketConn := *NewWebsocketConnection(
+			conn,
+			connId,
+			r.logger,
+			func() {
+				r.logger.Info().Str("auth_token", string(authToken)).Str("conn_id", string(connId)).Msg("removing subscriber websocket")
```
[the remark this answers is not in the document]
harryrackmil: yea good point - moved this to log the broker url instead since that should be unique to a given connection

cli/stork-publisher-agent/runner.go
```diff
@@ -0,0 +1,281 @@
...
+
+		conn, _, err := websocket.DefaultDialer.Dial(string(url), headers)
+		if err != nil {
+			r.logger.Error().Err(err).Msgf("Failed to connect to outgoing WebSocket: %v", err)
+			break
+		}
+
+		connId := ConnectionId(uuid.New().String())
+
+		r.logger.Info().Str("auth_token", string(authToken)).Str("conn_id", string(connId)).Msg("adding subscriber websocket")
```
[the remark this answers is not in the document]
harryrackmil: did a little cleanup here - removed connection id for the outgoing websocket, renamed subscriber -> receiver and changed the fields on log messages

cli/stork-publisher-agent/model.go
```diff
@@ -0,0 +1,118 @@
...
+	PublisherKey      string
+	SignatureType     string
+	QuantizedPrice    string
+)
+
+const EvmSignatureType = SignatureType("evm")
+const StarkSignatureType = SignatureType("stark")
+
+// Incoming
+type PriceUpdatePullWebsocket struct {
```
harryrackmil, line 40: Using different input objects for the pull-based websocket and the push based websocket. The data we're currently pulling has a `p` field so it felt weird to call that field Value.

It also seems like we can better avoid precision loss by converting strings to big.Float than by converting floats to big.Float, so for the push based message I'm using a string type rather than float.

akawalsky (commented):
Looks great, just a couple more comments

cli/stork-publisher-agent.go
```diff
@@ -0,0 +1,260 @@
...
+}
+
+// required
+const SignatureTypesFlag = "signature-types"
+const OracleIdFlag = "oracle-id"
+const StorkAuthFlag = "stork-auth"
+
+// optional
+const EvmPrivateKeyFlag = "evm-private-key"
+const EvmPublicKeyFlag = "evm-public-key"
```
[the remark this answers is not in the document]
harryrackmil: I tried pulling the stork public key out of the stork private key in a few ways but none of them seemed to work correctly - going to keep forcing users to specify the public key for now. Should be less of a pain once we're using the config file version of the publisher agent.

cli/stork-publisher-agent/processor.go
```diff
@@ -0,0 +1,188 @@
...
+	for val := range queue {
+		var priceUpdates []PriceUpdateWithTrigger
+		switch msg := val.(type) {
+		case DeltaTick:
+			priceUpdates = p.DeltaUpdate()
+		case ClockTick:
+			priceUpdates = p.ClockUpdate()
+		case PriceUpdate:
+			if p.signEveryUpdate {
+				priceUpdatesToSignCh <- PriceUpdateWithTrigger{PriceUpdate: msg, TriggerType: DeltaTriggerType}
```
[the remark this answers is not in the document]
harryrackmil: Confirmed that the Trigger field doesn't server any specific purpose beyond visibility - made a change so any update in "signEveryUpdate" mode has as its trigger type `unspecified` - we don't know what logic the publisher is doing behind the scenes.

cli/stork-publisher-agent/model.go
```diff
@@ -0,0 +1,120 @@
...
+	PublishTimestamp int64   `json:"t"`
+	Asset            AssetId `json:"a"`
+	Price            float64 `json:"p"`
+}
+
+type ValueUpdatePushWebsocket struct {
+	PublishTimestamp int64   `json:"t"`
+	Asset            AssetId `json:"a"`
+	ValueString      string  `json:"vs"`
+	ValueFloat       float64 `json:"vf"`
```
harryrackmil, line 50: we now allow users to pass either a string value or a float value - the string will have no precision loss but could be a little slower, while the float will be fast with a small amount of precision loss. A string that doesn't actually parse to a float will send an error response via the websocket.

cli/stork-publisher-agent/websocket.go
```diff
@@ -0,0 +1,421 @@
...
+
+func convertToValueUpdate(valueUpdatePushWebsocket ValueUpdatePushWebsocket) (*ValueUpdate, error) {
+	var value *big.Float
+	if len(valueUpdatePushWebsocket.ValueString) > 0 {
+		var success bool
+		value, success = new(big.Float).SetString(valueUpdatePushWebsocket.ValueString)
+		if !success {
+			return nil, errors.New("failed to convert string to float")
+		}
+	} else if valueUpdatePushWebsocket.ValueFloat > 0 {
```
harryrackmil, line 83: I want to explicitly fail if a user doesn't pass us either of the value types (rather than just treating it as value zero). This means that we can't actually receive a value of zero though. 

The most correct way to do this would be to include a binary or some kind of "type" indicator to make it clear which field we should look at, but that feels like unnecessary overhead if it's once per message

akawalsky (dismissed):

akawalsky (approved):

PR #7, opened 2024-08-28
Use Stork Registry in Publisher Agent

harryrackmil:
## Change
Hit the Stork Registry hosted on the Stork Rest API  every 10 minutes to update which signed prices should be sent to which brokers.

## Testing
* all data comes through when we configure both stark and evm pubs to use assets `*`
* if we delete a `PublisherConnection` between a publisher and a broker in dynamo, within 10 minutes we see no data from that publisher in the aggregator
* if we restrict a `PublisherConnection` to only BTCUSD, all other symbols stop reporting to the aggregator
* if we recreate the deleted `PublisherConnection` and restore the filter connection's assets to `*`, we see all data again in the aggregator

### DIFF
--- a/cli/stork-publisher-agent.go
+++ b/cli/stork-publisher-agent.go
@@ -25,6 +25,7 @@ const SignatureTypeFlag = "signature-type"
 const OracleIdFlag = "oracle-id"
 const PrivateKeyFlag = "private-key"
 const PublicKeyFlag = "public-key"
+const StorkAuthFlag = "stork-auth"
 
 // optional
 const ClockUpdatePeriodFlag = "clock-update-period"
@@ -32,7 +33,8 @@ const DeltaUpdatePeriodFlag = "delta-update-period"
 const ChangeThresholdPercentFlag = "change-threshold-percent"
 const IncomingWsPortFlag = "incoming-ws-port"
 const StorkRegistryBaseUrlFlag = "stork-registry-base-url"
-const StorkAuthFlag = "stork-auth"
+const StorkRegistryRefreshIntervalFlag = "stork-registry-refresh-interval"
+const BrokerReconnectDelayFlag = "broker-reconnect-delay"
 
 const PullBasedWsUrlFlag = "pull-based-ws-url"
 const PullBasedAuthFlag = "pull-based-auth"
@@ -48,18 +50,20 @@ func init() {
 	publisherAgentCmd.Flags().StringP(OracleIdFlag, "o", "", "oracle id (must be 5 characters)")
 	publisherAgentCmd.Flags().StringP(PrivateKeyFlag, "p", "", "Your private key for signing updates")
 	publisherAgentCmd.Flags().StringP(PublicKeyFlag, "k", "", "Your public key for signing updates")
+	publisherAgentCmd.Flags().StringP(StorkAuthFlag, "", "", "The auth token for Stork broker servers")
 
 	publisherAgentCmd.Flags().StringP(ClockUpdatePeriodFlag, "c", "500ms", "How frequently to update the price even if it's not changing much")
 	publisherAgentCmd.Flags().StringP(DeltaUpdatePeriodFlag, "d", "10ms", "How frequently to check if we're hitting the change threshold")
 	publisherAgentCmd.Flags().Float64P(ChangeThresholdPercentFlag, "t", 0.1, "Report prices immediately if they've changed by more than this percentage (1 means 1%)")
 	publisherAgentCmd.Flags().IntP(IncomingWsPortFlag, "i", 5215, "The port which you'll report prices to")
 	publisherAgentCmd.Flags().StringP(StorkRegistryBaseUrlFlag, "", ProdStorkRegistryBaseUrl, "The base URL for the Stork Registry (defaults to the production Stork Registry)")
-	publisherAgentCmd.Flags().StringP(StorkAuthFlag, "", "", "The auth token for Stork broker servers")
+	publisherAgentCmd.Flags().StringP(StorkRegistryRefreshIntervalFlag, "", "10m", "How frequently to refresh brokers from the Stork Registry")
+	publisherAgentCmd.Flags().StringP(BrokerReconnectDelayFlag, "", "5s", "The time to wait before reconnecting to a broker websocket after a failure")
 
 	publisherAgentCmd.Flags().StringP(PullBasedWsUrlFlag, "u", "", "A websocket URL to pull price updates from")
 	publisherAgentCmd.Flags().StringP(PullBasedAuthFlag, "a", "", "A Basic auth token needed to connect to the pull websocket")
-	publisherAgentCmd.Flags().StringP(PullBasedSubscriptionRequestFlag, "x", "", "A Basic auth token needed to connect to the pull websocket")
-	publisherAgentCmd.Flags().StringP(PullBasedReconnectDelayFlag, "r", "5s", "A Basic auth token needed to connect to the pull websocket")
+	publisherAgentCmd.Flags().StringP(PullBasedSubscriptionRequestFlag, "x", "", "A subscription message for the pull websocket")
+	publisherAgentCmd.Flags().StringP(PullBasedReconnectDelayFlag, "r", "5s", "The time to wait before reconnecting to the pull websocket after a failure")
 
 	publisherAgentCmd.Flags().BoolP(SignEveryUpdateFlag, "b", false, "Just sign every update received without any extra logic")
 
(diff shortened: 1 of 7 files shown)


cli/stork-publisher-agent.go
```diff
@@ -31,8 +31,8 @@ const ClockUpdatePeriodFlag = "clock-update-period"
 const DeltaUpdatePeriodFlag = "delta-update-period"
 const ChangeThresholdPercentFlag = "change-threshold-percent"
 const IncomingWsPortFlag = "incoming-ws-port"
-const OutgoingWsUrlFlag = "outgoing-ws-url"
-const OutgoingWsAuthFlag = "outgoing-ws-auth"
+const StorkRegistryBaseUrlFlag = "stork-registry-base-url"
```
harryrackmil, line 34: defaults to our prod API - only making it configurable so we can hit the dev registry

cli/stork-publisher-agent.go
```diff
@@ -31,8 +31,8 @@ const ClockUpdatePeriodFlag = "clock-update-period"
...
+const StorkAuthFlag = "stork-auth"
```
harryrackmil, line 35: used to auth into the stork registry API

cli/stork-publisher-agent/registry.go
```diff
@@ -0,0 +1,53 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get broker list: %v", err)
+	}
+	var brokers []BrokerConnectionConfig
+	err = json.Unmarshal(response, &brokers)
+	if err != nil {
+		return nil, fmt.Errorf("failed to unmarshal broker list: %v", err)
+	}
+
+	// combine all configs into a single asset map per url
```
harryrackmil, line 37: technically we could have multiple entries in the publisher connection table for a given Broker (since there could be multiple agreements with different payment terms for different assets). We merge all the results into a single map from broker URL to the set of assets for that URL

cli/stork-publisher-agent/rest.go
```diff
@@ -0,0 +1,43 @@
+package stork_publisher_agent
+
+import (
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+)
+
+func RestQuery(method string, baseUrl string, query url.Values, requestBody io.Reader, header http.Header) ([]byte, error) {
```
harryrackmil, line 10: copy pasted from stork-aggregator, added support for headers

cli/stork-publisher-agent/runner.go
```diff
@@ -15,7 +15,9 @@ type PublisherAgentRunner[T Signature] struct {
 	logger                  zerolog.Logger
 	priceUpdateCh           chan PriceUpdate
 	signedPriceBatchCh      chan SignedPriceUpdateBatch[T]
-	outgoingConnections     map[ConnectionId]*OutgoingWebsocketConnection[T]
+	registryClient          *RegistryClient
+	brokerMap               map[BrokerPublishUrl]map[AssetId]struct{}
```
harryrackmil, line 19: this stores the previously configured state of the world so we can see what needs to change when we get the latest configuration from the registry

cli/stork-publisher-agent/runner.go
```diff
@@ -45,6 +54,48 @@ func NewPublisherAgentRunner[T Signature](
...
+		r.logger.Error().Err(err).Msg("failed to get broker connections from Stork Registry")
+		return
+	}
+
+	// add or update desired connections
+	for brokerUrl, newAssetIdMap := range newBrokerMap {
+		_, exists := r.brokerMap[brokerUrl]
+		if exists {
+			// update connection
+			r.outgoingConnections[brokerUrl].UpdateAssets(newAssetIdMap)
```
harryrackmil, line 72: if the connection already exists, we overwrite the previously configured assets with the current configuration (this is just a cheap in-memory change so we don't check to see if it's changed or not)

cli/stork-publisher-agent/runner.go
```diff
@@ -45,6 +54,48 @@ func NewPublisherAgentRunner[T Signature](
...
+			r.outgoingConnections[url].Remove()
+			delete(r.brokerMap, url)
+		}
+	}
+
+	r.logger.Info().Msg("Broker connection updater finished")
+}
+
+func (r *PublisherAgentRunner[T]) RunBrokerConnectionUpdater() {
+	r.UpdateBrokerConnections()
```
harryrackmil, line 93: update once immediately, then every 10 minutes

cli/stork-publisher-agent/websocket.go
```diff
@@ -117,18 +118,34 @@ func (iwc *IncomingWebsocketConnection) Reader(priceUpdateCh chan PriceUpdate) {
...
 	}
 }
 
+func (owc *OutgoingWebsocketConnection[T]) UpdateAssets(assetIds map[AssetId]struct{}) {
+	owc.assetIdsLock.Lock()
+	owc.assetIds = assetIds
+	owc.assetIdsLock.Unlock()
+}
+
+func (owc *OutgoingWebsocketConnection[T]) Remove() {
```
harryrackmil, line 143: The websocket may be `Closed` if there's an error, after which we'll reconnect. If the broker has been removed from the registry, we want to exit the read loop and not reconnect, so we'll call `Remove()`

cli/stork-publisher-agent/websocket.go
```diff
@@ -149,8 +166,24 @@ func (owc *OutgoingWebsocketConnection[T]) Writer() {
 				logger.Warn().Msg("attempted to send message on closed websocket")
 				return
 			}
+
+			filteredPriceUpdates := make(SignedPriceUpdateBatch[T])
+			owc.assetIdsLock.RLock()
+			_, allAssets := owc.assetIds[WildcardSubscriptionAsset]
+			if allAssets {
+				filteredPriceUpdates = signedPriceUpdateBatch
```
harryrackmil, line 174: This is where the asset filtering happens - if the configuration changes in the registry we'll update assetIds and the filtering will happen without needing to touch the existing connection

cli/stork-publisher-agent/registry.go
```diff
@@ -0,0 +1,53 @@
...
+		baseUrl:   baseUrl,
+		authToken: authToken,
+	}
+}
+
+func (c *RegistryClient) GetBrokersForPublisher(publisherKey PublisherKey) (map[BrokerPublishUrl]map[AssetId]struct{}, error) {
+	brokerEndpoint := c.baseUrl + "/v1/registry/brokers"
+	queryParams := url.Values{}
+	queryParams.Add("publisher_key", string(publisherKey))
+	authHeader := http.Header{"Authorization": []string{"Basic " + string(c.authToken)}}
```
harryrackmil, line 26: this is kind of a cheat - we know that the auth token configured for use in the Stork Registry will also work to auth into the Stork Aggregator server.

Eventually this won't be the case, but we'll put the burden of authenticating public keys against the registry on the brokers, and we'll stop requiring auth tokens.

cli/stork-publisher-agent/runner.go
```diff
@@ -45,6 +54,48 @@ func NewPublisherAgentRunner[T Signature](
...
+	for range time.Tick(10 * time.Minute) {
```
akawalsky, line 94: Lets make this a configuration that defaults to 10 minutes

cli/stork-publisher-agent/runner.go
```diff
@@ -173,21 +222,27 @@ func (r *PublisherAgentRunner[T]) RunOutgoingConnection(url string, authToken Au
...
 		outgoingWebsocketConn.Writer()
-		r.logger.Warn().Msg("Outgoing websocket writer thread failed - reconnecting in 5 seconds")
-		time.Sleep(5 * time.Second)
+
+		if outgoingWebsocketConn.removed {
+			r.logger.Info().Msg("Outgoing websocket was removed - not reconnecting")
+			return
+		} else {
+			r.logger.Warn().Msg("Outgoing websocket writer thread failed - reconnecting in 5 seconds")
+			time.Sleep(5 * time.Second)
```
akawalsky, line 244: It also might make sense to make this a configuration with a default

cli/stork-publisher-agent/websocket.go
```diff
@@ -149,8 +166,24 @@ func (owc *OutgoingWebsocketConnection[T]) Writer() {
...
+			} else {
+				for asset, signedPriceUpdate := range signedPriceUpdateBatch {
```
akawalsky, line 176: It may be a bit of an over-optimization for now, but worth noting that since the expensive operation happening here is the signing, it is possible that some work is wasted if there are no consumers of the data for a given asset.

akawalsky (commented):
Just a couple comments, looks good!

cli/stork-publisher-agent/websocket.go
harryrackmil: ah yea that's a good point - I spent a few minutes trying to restrict the signing here but it involves some locking and information passing between objects that seems a little too complicated.

Going to keep signing everything we take in for now - in general I would assume a publisher agent will only be publishing data that someone somewhere is using, and if someone starts using a publisher agent's data for the first time I don't think the publisher would want their CPU usage to spike up suddenly. We can have predictable, sometimes inefficient processing for now and add this optimization if we find there's a need.

PR #17, opened 2024-09-12
[Publisher Agent] Add script to generate secrets.json file

harryrackmil:
## Change
* generate the secrets.json file for a publisher agent
* creates new EVM or Stark keys

## Testing
generated a secret file using this script and used it to start a publisher agent successfully


akawalsky (approved):
Would be cool to eventually fold this into the cli, looks good though

PR #21, opened 2024-09-25
[Publisher Agent] Reconnect if the Broker Dies

harryrackmil:
# Change
We observed a case where the Broker logged a read timeout from the publisher agent and the publisher agent never reconnected, so the publisher agent's data stopped reaching the broker.

I reproduced this situation locally by setting the broker read timeout to 10 nanoseconds and triggering a disconnection, but the publisher agent correctly reconnected to the broker.

It may be the case that something else (like a network outage) caused the publisher agent to not be able to reach the broker for the duration of the read timeout, and meant that the publisher agent never got a close message and therefore never attempted to reconnect.

As a blanket fix, we add a write timeout to the publisher agent so that we'll attempt to reconnect to any broker if it takes more than 10 seconds to write to it. We'll also remove an unnecessary `break` from the reconnect loop - we should really never exit that loop unless we've deregistered the broker.

# Testing
Before making these changes, I ran the publisher agent and broker locally and then stopped the broker container, and observed that the publisher agent just hung infinitely and didn't log any errors or attempt to reconnect.

After these changes, the publisher agent attempts to reconnect every 5 seconds and when I start the broker container up again it resumes publishing normally.

### DIFF
--- a/cli/stork-publisher-agent/runner.go
+++ b/cli/stork-publisher-agent/runner.go
@@ -10,16 +10,16 @@ import (
 )
 
 type PublisherAgentRunner[T Signature] struct {
-	config                  StorkPublisherAgentConfig
-	signatureType           SignatureType
-	logger                  zerolog.Logger
-	ValueUpdateCh           chan ValueUpdate
-	signedPriceBatchCh      chan SignedPriceUpdateBatch[T]
-	registryClient          *RegistryClient
-	brokerMap               map[BrokerPublishUrl]map[AssetId]struct{}
-	outgoingConnections     map[BrokerPublishUrl]*OutgoingWebsocketConnection[T]
-	outgoingConnectionsLock sync.RWMutex
-	signer                  Signer[T]
+	config                      StorkPublisherAgentConfig
+	signatureType               SignatureType
+	logger                      zerolog.Logger
+	ValueUpdateCh               chan ValueUpdate
+	signedPriceBatchCh          chan SignedPriceUpdateBatch[T]
+	registryClient              *RegistryClient
+	assetsByBroker              map[BrokerPublishUrl]map[AssetId]struct{}
+	outgoingConnectionsByBroker map[BrokerPublishUrl]*OutgoingWebsocketConnection[T]
+	outgoingConnectionsLock     sync.RWMutex
+	signer                      Signer[T]
 }
 
 func NewPublisherAgentRunner[T Signature](
@@ -43,7 +43,7 @@ func NewPublisherAgentRunner[T Signature](
 		ValueUpdateCh:           make(chan ValueUpdate, 4096),
 		signedPriceBatchCh:      make(chan SignedPriceUpdateBatch[T], 4096),
 		registryClient:          registryClient,
-		brokerMap:               make(map[BrokerPublishUrl]map[AssetId]struct{}),
+		assetsByBroker:          make(map[BrokerPublishUrl]map[AssetId]struct{}),
 		outgoingConnections:     make(map[BrokerPublishUrl]*OutgoingWebsocketConnection[T]),
 		outgoingConnectionsLock: sync.RWMutex{},
 		signer:                  *signer,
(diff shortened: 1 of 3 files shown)


cli/stork-publisher-agent/runner.go
```diff
@@ -76,10 +76,12 @@ func (r *PublisherAgentRunner[T]) UpdateBrokerConnections() {
 
 	// add or update desired connections
 	for brokerUrl, newAssetIdMap := range newBrokerMap {
-		_, exists := r.brokerMap[brokerUrl]
-		if exists {
+		_, brokerExists := r.brokerMap[brokerUrl]
+		outgoingConnection, outgoingConnectionExists := r.outgoingConnections[brokerUrl]
```
harryrackmil, line 80: got a nil reference error when I was trying to break things and refreshing the registry really frequently - if an outgoing connection object is missing we should create it.

cli/stork-publisher-agent/runner.go
```diff
@@ -147,7 +149,8 @@ func (r *PublisherAgentRunner[T]) RunOutgoingConnection(url BrokerPublishUrl, as
 		conn, _, err := websocket.DefaultDialer.Dial(string(url), headers)
 		if err != nil {
 			r.logger.Error().Err(err).Msgf("Failed to connect to outgoing WebSocket: %v", err)
-			break
+			time.Sleep(r.config.BrokerReconnectDelay)
+			continue
```
harryrackmil, line 153: if we fail to connect to a broker, we should try again in 5 seconds.

cli/stork-publisher-agent/websocket.go
```diff
@@ -312,10 +313,15 @@ func SendWebsocketMsg[T any](conn *websocket.Conn, msgType string, data T, trace
...
-	return sendWebsocketResponse[WebsocketMessage[T]](conn, msg, logger)
+	return sendWebsocketResponse[WebsocketMessage[T]](conn, msg, logger, OutgoingWriteTimeout)
 }
 
-func sendWebsocketResponse[T any](conn *websocket.Conn, msg T, logger zerolog.Logger) error {
+func sendWebsocketResponse[T any](conn *websocket.Conn, msg T, logger zerolog.Logger, writeTimeout time.Duration) error {
+	if writeTimeout.Nanoseconds() > 0 {
+		deadline := time.Now().Add(writeTimeout)
+		_ = conn.SetWriteDeadline(deadline)
+	}
```
harryrackmil, line 323: set a write deadline so we get a timeout if the broker isn't reachable

cli/stork-publisher-agent/signer_test.go
```diff
@@ -28,6 +28,7 @@ func TestSigner_GetSignedPriceUpdate_Evm(t *testing.T) {
 		"",
 		"",
 		time.Duration(0),
+		time.Duration(0),
```
harryrackmil, line 31: fix tests

cli/stork-publisher-agent/runner.go
```diff
@@ -10,16 +10,16 @@ import (
...
-	outgoingConnectionsLock sync.RWMutex
-	signer                  Signer[T]
+	config                      StorkPublisherAgentConfig
+	signatureType               SignatureType
+	logger                      zerolog.Logger
+	ValueUpdateCh               chan ValueUpdate
+	signedPriceBatchCh          chan SignedPriceUpdateBatch[T]
+	registryClient              *RegistryClient
+	assetsByBroker              map[BrokerPublishUrl]map[AssetId]struct{}
+	outgoingConnectionsByBroker map[BrokerPublishUrl]*OutgoingWebsocketConnection[T]
```
harryrackmil, line 20: renamed for clarity - one map stores the connection object and the other stores the assets we want to send through that connection.

cli/stork-publisher-agent/runner.go
```diff
@@ -76,23 +76,23 @@ func (r *PublisherAgentRunner[T]) UpdateBrokerConnections() {
 
 	// add or update desired connections
 	for brokerUrl, newAssetIdMap := range newBrokerMap {
-		_, exists := r.brokerMap[brokerUrl]
-		if exists {
+		outgoingConnection, outgoingConnectionExists := r.outgoingConnectionsByBroker[brokerUrl]
```
harryrackmil, line 79: the old logic here was kind of weird - we really only want to update connections which still exist, regardless of whether we have an existing entry in the broker asset map. 

We now only look at the connection map when deciding whether to create a new connection, and the asset map is basically a configuration for each connection.

akawalsky (dismissed):

PR #22, opened 2024-09-27
Remove break from puller reconnect loop

harryrackmil:
# Change
Right now if we fail to connect to the pull-based websocket server we'll exit the for loop and never try to reconnect. We should instead just wait 5 seconds and reconnect like we do for other types of errors when reading from the pull-based websocket


harryrackmil:
> Im seeing that we still call `break` in the nested for loop. Do we want to continue to do that?

Yes I believe that's correct - the inner loop iterates over messages, and if we get a bad message or an error on read we want to stop reading and reconnect, so we're breaking out of the inner loop and reconnecting as desired.

PR #23, opened 2024-09-27
Reorganize external repo

akawalsky:


apps/README.md
```diff
@@ -0,0 +1,19 @@
+# Stork Apps
+
+A suite of tools to interact with Stork's services and on-chain contracts.
+
+## pusher
```
harryrackmil, line 5: nit - do we want to use `Pusher` and `Publisher Agent` instead of the snake case versions for the readme titles? Fine with this if it's a convention or something

README.md
```diff
@@ -8,7 +8,7 @@ Stork leverages cryptographic primitives used in blockchain technologies to secu
 
 This repository contains various components utilized by data publishers and subscribers to leverage Stork's technology.
 
-#### CLI
+#### Apps
```
harryrackmil, line 11: nit - should we link to the apps Readme in this section?

apps/cmd/pusher/main.go
```diff
@@ -8,14 +8,16 @@ import (
...
+	"github.com/Stork-Oracle/stork_external/lib/pusher"
 )
 
 var verbose bool
 
 func main() {
 	rootCmd := &cobra.Command{
-		Use:   "stork",
-		Short: "Stork CLI tool for EVM contract updates based on WebSocket prices",
+		Use:   "stork-push",
```
harryrackmil, line 19: how do you feel about calling this app something like `chain_pusher` and making this command `stork-chain-pusher`? Don't feel super strongly about it, but `stork-push` is a little general - sounds like it could be pushing into stork too.

apps/docs/pusher.md
```diff
@@ -1,16 +1,4 @@
...
-## evm-push
-
-### Background
-
-Stork signed data feeds are delivered off-chain from publishers to subscribers via Stork's aggregation network. In order for this data to be usable on-chain, it must be written to the Stork contract on any EVM compatible network. This tool is used to push signed data feeds to the Stork contract.
-
-Stork does not write this data to the chain directly by default, but any subscriber can choose to write the data to the chain if they so choose. This tool can be used to facilitate that process.
-
-### Usage
+## Usage
```
harryrackmil, line 1: should probably still have some kinda top level header like `# Pusher` or `# Chain Pusher`

apps/pusher.Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+COPY lib/ ./lib/
+COPY cmd/pusher/ ./cmd/pusher/
+
+ARG TARGETPLATFORM
+RUN case "$TARGETPLATFORM" in \
+        "linux/amd64")  export GOARCH="amd64" && export CC="x86_64-linux-gnu-gcc";; \
+        "linux/arm64")  export GOARCH="arm64" && export CC="aarch64-linux-gnu-gcc";; \
+        *) echo "Unsupported platform: $TARGETPLATFORM" ; exit 1 ;; \
+    esac && \
+    CGO_ENABLED=1 GOOS=linux GOARCH=$GOARCH CC=$CC go build -o /stork cmd/pusher/main.go
```
harryrackmil, line 26: do we need to do cross-platform compiling for the pusher?

apps/scripts/push_to_docker.sh
```diff
@@ -2,9 +2,16 @@
 
 set -e
 
+IMAGE_NAME=$1
+
+if [ -z "$IMAGE_NAME" ]; then
```
harryrackmil, line 7: nit - I think `set -u` will accomplish the same thing - it will fail the script if $1 is unset. May fail with a less helpful error though so this makes sense too

harryrackmil (approved):

apps/pusher.Dockerfile
akawalsky: I don't see why not. We would prob like for anyone to be able to run this as well

PR #25, opened 2024-10-01
Allow publishing only on price change

harryrackmil:
# Change
There was a bug which meant that users who didn't want to use clock updates or sign every update would never actually publish a price, even if there were a large delta. 

This is because we had made an assumption that either the clock logic or the "sign every update" logic would trigger an initial update, and the delta update logic would compare current price to the last sent update. But if we only use delta logic to update, we'll never send an initial price to compare to, meaning we'll never send an update.

Instead, the delta checker should send an initial update if nothing has been sent, and after that update it will have a point of comparison for future updates.

# Testing
[X] Added a unit test
[X] ran locally confirmed the broker receives results

### DIFF
--- a/cli/stork-publisher-agent/processor_test.go
+++ b/cli/stork-publisher-agent/processor_test.go
@@ -1,0 +1,115 @@
+package stork_publisher_agent
+
+import (
+	"math/big"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/rs/zerolog"
+	"github.com/stretchr/testify/assert"
+)
+
+const evmPrivateKey = "0x8b558d5fc31eb64bb51d44b4b28658180e96764d5d5ac68e6d124f86f576d9de"
+const evmPublicKey = "0x99e295e85cb07c16b7bb62a44df532a7f2620237"
+const starkPublicKey = "0x66253bdeb3c1a235cf4376611e3a14474e2c00fd2fb225f9a388faae7fb095a"
+const starkPrivateKey = "0x418d3fd8219a2cf32a00d458f61802d17f01c5bcde5a4f82008ee4a7c8e9a06"
+const storkAuth = "fake_auth"
+
+const assetId = "fakeAsset"
+
+func getNextSignedOutput[T Signature](ch chan SignedPriceUpdateBatch[T], timeout time.Duration) (SignedPriceUpdateBatch[T], bool) {
+	select {
+	case value := <-ch:
+		return value, true
+	case <-time.After(timeout):
+		return nil, false
+	}
+}
+
+func TestDeltaOnly(t *testing.T) {
+	logger := zerolog.New(os.Stdout).With().Timestamp().Logger()
+	config := NewStorkPublisherAgentConfig(
+		[]SignatureType{EvmSignatureType},
+		evmPrivateKey,
+		evmPublicKey,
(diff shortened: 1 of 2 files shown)


cli/stork-publisher-agent/processor_test.go
```diff
@@ -0,0 +1,115 @@
...
+
+const evmPrivateKey = "0x8b558d5fc31eb64bb51d44b4b28658180e96764d5d5ac68e6d124f86f576d9de"
+const evmPublicKey = "0x99e295e85cb07c16b7bb62a44df532a7f2620237"
+const starkPublicKey = "0x66253bdeb3c1a235cf4376611e3a14474e2c00fd2fb225f9a388faae7fb095a"
+const starkPrivateKey = "0x418d3fd8219a2cf32a00d458f61802d17f01c5bcde5a4f82008ee4a7c8e9a06"
+const storkAuth = "fake_auth"
+
+const assetId = "fakeAsset"
+
+func getNextSignedOutput[T Signature](ch chan SignedPriceUpdateBatch[T], timeout time.Duration) (SignedPriceUpdateBatch[T], bool) {
```
harryrackmil, line 21: helper function to get the next signed price, but time out after some amount of time so we don't wait infinitely in the failure case.

cli/stork-publisher-agent/processor_test.go
```diff
@@ -0,0 +1,115 @@
...
+func TestDeltaOnly(t *testing.T) {
+	logger := zerolog.New(os.Stdout).With().Timestamp().Logger()
+	config := NewStorkPublisherAgentConfig(
+		[]SignatureType{EvmSignatureType},
+		evmPrivateKey,
+		evmPublicKey,
+		starkPrivateKey,
+		starkPublicKey,
+		time.Duration(0),
+		10*time.Millisecond,
```
harryrackmil, line 39: ClockPeriod is set to zero and SignEveryUpdate is false, so only the delta logic will run.

cli/stork-publisher-agent/processor_test.go
```diff
@@ -0,0 +1,115 @@
...
+		0,
+	)
+
+	signer, err := NewSigner[*EvmSignature](*config, EvmSignatureType, logger)
+	if err != nil {
+		t.Fatalf("NewSigner[*EvmSignature]: %v", err)
+	}
+
+	inputCh := make(chan ValueUpdate)
+	outputCh := make(chan SignedPriceUpdateBatch[*EvmSignature])
```
harryrackmil, line 61: To test the processor logic we'll create the input channel (value updates) and output channel (signed updates) and make sure that the processor is correctly deciding when to send out a signed update.

cli/stork-publisher-agent/processor_test.go
```diff
@@ -0,0 +1,115 @@
...
+		DefaultChangeThresholdPercent,
+		false,
+		inputCh,
+		outputCh,
+		logger,
+	)
+
+	go processor.Run()
+
+	// initial update gets sent out
```
harryrackmil, line 76: test confirms that:
* first update gets signed and sent out
* subsequent update with no price change doesn't get signed or sent
* subsequent update with significant price change gets signed and sent out

akawalsky (approved):

PR #26, opened 2024-10-02
Add test for zero price case

harryrackmil:
Make sure that zero prices are handled correctly in delta calculation (even though we're dividing by zero)


cli/stork-publisher-agent/processor_test.go
```diff
@@ -113,3 +113,107 @@ func TestDeltaOnly(t *testing.T) {
...
+		DefaultChangeThresholdPercent,
+		false,
+		inputCh,
+		outputCh,
+		logger,
+	)
+
+	go processor.Run()
+
+	// initial zero update gets sent out
```
harryrackmil, line 163: 1. initial zero price update triggers the delta
2. subsequent zero price update does not
3. nonzero price update triggers the delta
4. returning to zero triggers the delta

akawalsky (approved):

PR #28, opened 2024-10-03
Clearer configuration error reporting

harryrackmil:
# Change
Log more clearly for various types of configuration issues.

# Testing
If a user doesn't configure the stork auth, or configures a badly formatted auth like `username:password`, the publisher agent will exit with this log message:
```
Error: error loading config: stork auth token must a non-empty base64 string
```

If a user doesn't configure the private or public key for their configured signature type or passes a badly formatted key, the agent will exit with a log message like:
```
Error: error loading config: must pass a valid EVM public key
```

If the stork auth is well formed but not valid, the publisher will log this error (but not fail):
```
{"level":"error","application":"stork-publisher-agent","service":"runner","signature_type":"evm","error":"not authorized to query Stork Registry - check your configured StorkAuth","time":"2024-10-03T01:50:30.093097305Z","message":"failed to get broker connections from Stork Registry"}
```

If the user's public key is not in the Stork Registry we'll log this error (but not fail):
```
{"level":"warn","application":"stork-publisher-agent","service":"runner","signature_type":"evm","time":"2024-10-03T02:14:44.177937339Z","message":"no stork registry broker found for public key (0xEE992A2493bcf5f68A3cA395109c0dc02333FF85) - reach out to Stork to make sure you're whitelisted"}
```

### DIFF
--- a/apps/README.md
+++ b/apps/README.md
@@ -8,7 +8,7 @@ Stork signed data feeds are delivered off-chain from publishers to subscribers v
 
 Stork does not write this data to the chain directly by default, but any subscriber can choose to write the data to the chain if they so choose. This tool can be used to facilitate that process.
 
-See [Stork Pusher Docs](docs/chain_pusher).
+See [Stork Pusher Docs](docs/chain_pusher.md).
 
 ## Publisher Agent
 
@@ -16,4 +16,4 @@ The Stork Network receives signed data feeds from publishers and aggregates them
 
 The easiest way to become a Stork Publisher is to run the Stork Publisher Agent docker container on your infrastructure and send price updates to the Agent through a local websocket. The Stork Publisher Agent will sign your price updates with your private key and send them to the Stork Network.
 
-See [Stork Publisher Agent Docs](docs/publisher_agent).
+See [Stork Publisher Agent Docs](docs/publisher_agent.md).
--- a/apps/lib/publisher_agent/config.go
+++ b/apps/lib/publisher_agent/config.go
@@ -13,6 +13,7 @@ import (
 )
 
 var Hex32Regex = regexp.MustCompile(`^0x[0-9a-fA-F]+$`)
+var StorkAuthRegex = regexp.MustCompile(`^[0-9a-fA-F]+$`)
 
 const DefaultClockUpdatePeriod = "500ms"
 const DefaultDeltaUpdatePeriod = "10ms"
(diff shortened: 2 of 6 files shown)


apps/README.md
```diff
@@ -8,12 +8,12 @@ Stork signed data feeds are delivered off-chain from publishers to subscribers v
 
 Stork does not write this data to the chain directly by default, but any subscriber can choose to write the data to the chain if they so choose. This tool can be used to facilitate that process.
 
-See [Stork Pusher Docs](docs/chain_pusher).
+See [Stork Pusher Docs](docs/chain_pusher.md).
```
harryrackmil, line 11: these links were broken

apps/lib/publisher_agent/config.go
```diff
@@ -13,6 +13,7 @@ import (
 )
 
 var Hex32Regex = regexp.MustCompile(`^0x[0-9a-fA-F]+$`)
+var StorkAuthRegex = regexp.MustCompile(`^[A-Za-z0-9+/]+={0,2}$`)
```
harryrackmil, line 16: ensure it's base64 encoded

apps/lib/publisher_agent/registry.go
```diff
@@ -30,14 +33,31 @@ func (c *RegistryClient) GetBrokersForPublisher(publisherKey signer.PublisherKey
 	if err != nil {
 		return nil, fmt.Errorf("failed to get broker list: %v", err)
 	}
+
 	var brokers []BrokerConnectionConfig
 	err = json.Unmarshal(response, &brokers)
 	if err != nil {
-		return nil, fmt.Errorf("failed to unmarshal broker list: %v", err)
+		// check if the registry returned an error
+		var errorResponse RegistryErrorResponse
```
harryrackmil, line 41: the rest query might succeed, but return an error message. We should make clear the error is clear (since it currently says "failed to unmarshal broker list" when the stork auth is incorrect)

apps/lib/publisher_agent/registry.go
```diff
@@ -30,14 +33,31 @@ func (c *RegistryClient) GetBrokersForPublisher(publisherKey signer.PublisherKey
...
+				return nil, fmt.Errorf("failed to query the Stork Registry: %s", errorResponse.Error)
+			}
+		}
+		return nil, fmt.Errorf("failed to unmarshal response from Stork Registry: %s", string(response))
 	}
 
-	// combine all configs into a single asset map per url
 	brokerMap := make(map[BrokerPublishUrl]map[AssetId]struct{})
+
+	if len(brokers) == 0 {
```
harryrackmil, line 55: make it explicit if the agent is able to reach the registry, but isn't registered

akawalsky (approved):

PR #29, opened 2024-10-15
Add Stark Validation to Signer

harryrackmil:
# Change
* Add stark validation code in Rust
* Add stark and evm validation code in Go
* Add a docker image for the `signer` package
* Support importing the `signer` package into other go projects

# Testing
[X] signer can be used outside this repo
[X] publisher agent still works


apps/lib/signer/verifier_test.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+
+	// change the price slightly
+	err = VerifyStarkPublisherPrice(1708940577123456789, "0x44594458555344000000000000000000637a6f7778", "3335950348880000000", pubKey, signature)
+	if err == nil {
+		t.Fail()
+	}
+}
+
+func TestStarkVerifier_VerifyPublisherPriceLongAssetId(t *testing.T) {
```
harryrackmil, line 46: an earlier version of the validator was failing with assets with super long names (I was assuming the asset hex would have length 32 exactly) so testing this case specifically

publisher-agent.Dockerfile
```diff
@@ -1,35 +1,6 @@
...
-    g++-x86-64-linux-gnu libc6-dev-amd64-cross \
-    build-essential \
-    curl \
-    git \
-    bash
-
-
-ENV CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER=aarch64-linux-gnu-gcc
-ENV CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_LINKER=x86_64-linux-gnu-gcc
+FROM storknetwork/signer:v1.0.0-dev AS rust-build
```
harryrackmil, line 2: move this logic to the `signer` base image

harryrackmil, line 2: todo: use `v1.0.0` once this is ready to merge

apps/lib/signer/model.go
```diff
@@ -7,6 +7,7 @@ type (
 	EvmPrivateKey     string
 	StarkPublisherKey string
 	StarkPrivateKey   string
+	QuantizedPrice    string
```
akawalsky, line 10: When I refactored this I was somewhat hesitant to include this because it felt kind of application specific. Is there a reason you decided to include it?

apps/lib/signer/verifier_test.go
```diff
@@ -0,0 +1,57 @@
+package signer
+
+import (
+	"testing"
+)
+
+func TestEvmVerifier_VerifyPublisherPrice(t *testing.T) {
```
akawalsky, line 7: Lets throw in some failure case tests to be safe

akawalsky (approved):
Just a couple comments, looks great

akawalsky (dismissed):

PR #31, opened 2024-10-17
bump juno version

harryrackmil:
* bump the juno version to 11.7 so we don't depend on the private sourcegraph repo
* set min go version back to 1.22 for compatibility with stork-aggregator


go.mod
```diff
@@ -1,12 +1,12 @@
...
 
-go 1.22.2
+go 1.22.0
 
 require (
-	github.com/NethermindEth/juno v0.3.1
-	github.com/consensys/gnark-crypto v0.12.1
-	github.com/ethereum/go-ethereum v1.14.7
-	github.com/google/uuid v1.5.0
+	github.com/NethermindEth/juno v0.11.7
```
harryrackmil, line 6: I only changed this version and the `go 1.22` - the rest is from `go mod tidy`

akawalsky (approved):

PR #32, opened 2024-10-18
Allow Tracking Publisher Metadata

harryrackmil:
# Change
* Publishers will report their public ip, publisher agent version, architecture, AWS availability zone and AWS instance type to Stork's API hourly
* They can toggle this off by setting `PublisherMetadataUpdateInterval` to 0 in their config


apps/lib/publisher_agent/config.go
```diff
@@ -19,25 +19,27 @@ const DefaultClockUpdatePeriod = "500ms"
...
-	PullBasedWsSubscriptionRequest string
-	PullBasedWsReconnectDelay      string
-	PullBasedWsReadTimeout         string
-	SignEveryUpdate                bool
-	IncomingWsPort                 int
+	SignatureTypes                   []signer.SignatureType
+	ClockPeriod                      string
+	DeltaCheckPeriod                 string
+	ChangeThresholdPercent           float64 // 0-100
+	StorkRestApiBaseUrl              string
```
harryrackmil, line 33: renamed `StorkRegistryBaseUrl` -> `StorkRestApiBaseUrl` so we can use it for both the registry and metadata reporting.

apps/lib/publisher_agent/metadata.go
```diff
@@ -0,0 +1,153 @@
...
+		logger:              logger,
+	}
+}
+
+func (p *PublisherMetadataReporter) Run() {
+	err := p.report()
+	if err != nil {
+		p.logger.Warn().Msgf("Error reporting publisher metadata: %v", err)
+	}
+	for range time.Tick(p.reportPeriod) {
```
harryrackmil, line 67: it's maybe a little silly to push this metadata more than once since nothing is expected to change.

It's nice to be able to treat it as a heartbeat too though - if a particular key hasn't reported for an hour then we can assume the agent has stopped. Also useful if we add more dynamic metadata in the future.

apps/lib/publisher_agent/metadata.go
```diff
@@ -0,0 +1,153 @@
...
+		err = p.report()
+		if err != nil {
+			p.logger.Warn().Msgf("Error reporting publisher metadata: %v", err)
+		}
+	}
+}
+
+func (p *PublisherMetadataReporter) report() error {
+	metadata := p.getMetadata()
```
harryrackmil, line 76: similarly, we could probably pull metadata once and then just report the same thing many times. I'm doing it every time so we don't get into a situation where we fail to query something once and then can never have that metadata.

apps/lib/publisher_agent/metadata.go
```diff
@@ -0,0 +1,153 @@
...
+		return AwsMetadata{
+			IsAws: false,
+		}
+	}
+
+	metadata := AwsMetadata{
+		IsAws: true,
+	}
+
+	azResult, err := RestQuery("GET", awsMetadataUrl+"/placement/availability-zone-id", nil, nil, nil)
```
harryrackmil, line 142: using availability zone id (e.g. `apne1-az4`) instead of availability zone (`ap-northeast-1c`) since availability zone id is the same across all AWS accounts and availability zone isn't guaranteed to be.

apps/lib/publisher_agent/rest.go
```diff
@@ -26,7 +27,9 @@ func RestQuery(method string, baseUrl string, query url.Values, requestBody io.R
 		return nil, fmt.Errorf("error creating %s request for url %s: %v", method, urlString, err)
 	}
 
-	client := &http.Client{}
+	client := &http.Client{
+		Timeout: 5 * time.Second,
+	}
```
harryrackmil, line 32: all rest requests time out after 5 seconds

apps/lib/publisher_agent/runner.go
```diff
@@ -118,6 +130,10 @@ func (r *PublisherAgentRunner[T]) Run() {
 
 	go r.RunBrokerConnectionUpdater()
 
+	if r.config.PublisherMetadataUpdateInterval.Nanoseconds() > 0 {
```
harryrackmil, line 133: don't run the metadata updater if this configuration is set to `0ms`

version.txt
```diff
@@ -0,0 +1 @@
+v0.1.0-dev
```
harryrackmil, line 1: todo: update to `v0.1.0` once this is approved

akawalsky (approved):

PR #33, opened 2024-10-22
build: improve cgo flag management

asonawalla:
The hard-coded paths CFLAGS and LDFLAGS work for the docker build we have specified here, but cause problems with downstream imports which have a different directory structure.

This change:
- removes the hardcoded paths to keep only the relative ones
- consolidates flags into a single .go source file to avoid linker warnings about duplicates
- bonus: fix unused import warning from cargo

### DIFF
--- a/apps/lib/signer/cgo_flags.go
+++ b/apps/lib/signer/cgo_flags.go
@@ -1,0 +1,7 @@
+package signer
+
+/*
+#cgo LDFLAGS: -L./rust/stork/target/release -lstork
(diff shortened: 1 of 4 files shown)


asonawalla:
@harryrackmil I confirmed that we can still build the publisher-agent and signer containers after making this change for amd & arm architectures.

PTAL

asonawalla:
Thanks!  Mind merging?  I don't have write access on this repo.

PR #34, opened 2024-10-22
Allow secrets to be passed as env vars in publisher agent

HenryMBaldwin:
# Change

- Settings in the secrets.json can also be passed as environment variables
- Secrets.json is now optional (as is the -k flag)
- Updated docs to reflects this
- Made unrelated link fixes in docs

# Testing

- Unit testing patterned after config unit testing
- Recompiled docker image with new code locally and testing passing environment variables


apps/docs/publisher_agent.md
```diff
@@ -29,10 +29,29 @@ You can also generate your own keys and build your own `secrets.json` file - it
 }
 ```
 
-You can run the docker container like this:
+The secrets.json file can be subsituted or combined with environment variables if preferred. If the same property has a value set in both the secrets.json and environment variable, the environment variable take priority.
```
akawalsky, line 32:
```suggestion
The secrets.json file can be subsituted or combined with environment variables if preferred. If the same property has a value set in both the secrets.json and environment variable, the environment variable takes priority.
```

apps/lib/publisher_agent/config.go
```diff
@@ -251,6 +256,38 @@ func LoadConfig(configFilePath string, keysFilePath string) (*StorkPublisherAgen
 	return config, nil
 }
 
+// this overwrites any existing values in keysFile
+func LoadKeysFileFromEnv(keysFile *KeysFile) {
```
akawalsky, line 260: This prob doesn't need to be capitalized - capitalization is akin to marking a fn as public for the module

apps/docs/publisher_agent.md
```diff
@@ -9,7 +9,7 @@ To run the agent on your infrastructure, you'll need to first create a `config.j
 
 Here we open a websocket at `ws://localhost:5216/publish` to receive price updates. The agent will sign and send the latest price every 500 ms (clock updates) or when the price has changed by more than 0.1% (delta updates). 
 
-`config.json` follows the structure of the [ConfigFile struct](config.go):
+`config.json` follows the structure of the [ConfigFile struct](../lib/publisher_agent/config.go#L27):
```
harryrackmil, line 12: ah good catch. Let's leave out the line number though - otherwise we'll have to keep updating the README whenever we change this go file

apps/docs/publisher_agent.md
```diff
@@ -19,7 +19,7 @@ Here we open a websocket at `ws://localhost:5216/publish` to receive price updat
 
 Once you've gotten a StorkAuth key from Stork, you can use the [generate_secrets.py](../../python/src/generate_secrets/generate_secrets.py) script to generate your EVM and/or Stork keys and to build the `secrets.json` file. Make sure not to check this file into version control or share it in any way. 
 
-You can also generate your own keys and build your own `secrets.json` file - it follows the structure of the [KeysFile struct](config.go):
+You can also generate your own keys and build your own `secrets.json` file - it follows the structure of the [KeysFile struct](../lib/publisher_agent/config.go#L43):
```
harryrackmil, line 22: same here

apps/docs/publisher_agent.md
```diff
@@ -29,10 +29,29 @@ You can also generate your own keys and build your own `secrets.json` file - it
 }
 ```
 
-You can run the docker container like this:
+The secrets.json file can be subsituted or combined with environment variables if preferred. If the same property has a value set in both the secrets.json and environment variable, the environment variable take priority.
+
+The available environment variables are:
+| Key | Details |
+|---------|---------|
+| STORK_EVM_PRIVATE_KEY | [EVM Private Key](../lib/signer/model.go#L7) |
```
harryrackmil, line 37: same comment - maybe this can just be a list of keys without the "details" column

apps/lib/publisher_agent/config.go
```diff
@@ -86,11 +84,18 @@ func LoadConfig(configFilePath string, keysFilePath string) (*StorkPublisherAgen
 	}
 
 	var keysFile KeysFile
```
akawalsky, line 86: Can we change the name of this struct? I feel like instead of KeysFile it should be `Secrets` and the other struct should be `Config`

apps/lib/publisher_agent/config_test.go
```diff
@@ -21,3 +22,23 @@ func TestLoadConfig(t *testing.T) {
 	assert.Equal(t, "", config.PullBasedWsUrl)
 	assert.Equal(t, 5216, config.IncomingWsPort)
 }
+
+func TestLoadKeysFileFromEnv(t *testing.T) {
```
akawalsky, line 26: Can you add another test to ensure the cascading prioritization works as expected?

akawalsky (commented):
Nice! Just a couple comments here. @harryrackmil to review as well

apps/lib/publisher_agent/config.go
harryrackmil: I named it keys initially because some of the data isn't actually secret (e.g. public keys), but `secrets` is fine too. we should probably move the public keys into config eventually but let's hold off on that for now since users will need to update their config files

apps/lib/publisher_agent/config.go
```diff
@@ -86,11 +84,18 @@ func LoadConfig(configFilePath string, keysFilePath string) (*StorkPublisherAgen
...
-	err = json.Unmarshal(keyFileData, &keysFile)
-	if err != nil {
-		return nil, fmt.Errorf("failed to unmarshal keys file: %w", err)
+
+	// only deserialize keyFileData if keysFilePath was successfully read
+	if keyFileReadErr == nil {
```
harryrackmil, line 89: let's reorder this file a bit so we do all the loading for the config file followed by all the loading for the keys file.

feels a little weird to check for existence of an error a few lines later if we can avoid it - would be nice to just try to load the file and immediately take action on it

apps/lib/publisher_agent/config.go
```diff
@@ -86,11 +84,18 @@ func LoadConfig(configFilePath string, keysFilePath string) (*StorkPublisherAgen
...
+		err = json.Unmarshal(keyFileData, &keysFile)
+		if err != nil {
+			return nil, fmt.Errorf("failed to unmarshal keys file: %w", err)
+		}
 	}
 
+	// overwrite keysFile with any set env vars
+	loadKeysFileFromEnv(&keysFile)
```
harryrackmil, line 97: I generally don't like the pattern of passing a variable to a static function and having the function mutate that variable in place, since someone reading this code might not expect this variable to change after this call.

What do you think about making loadKeysFileFromEnv a method of the `KeysFile` object, so we would call something like `keysFile.updateFromEnvVars()` to make it a little more obvious that we're changing the object.

harryrackmil (commented):
looks good! Left some comments

apps/lib/publisher_agent/config.go
HenryMBaldwin: I agree, although there is still public info in there (pub keys, I assume oracle id is public as well). Should the public stuff be moved into config? Or are we okay with secrets not being only secrets?

HenryMBaldwin: https://github.com/Stork-Oracle/stork-external/pull/34#discussion_r1811269653

nvm, @harryrackmil actually already answered this. I'm fine to just do the renaming but keep the structure for now.

akawalsky: I'm more commenting on the notion of it being a "file" than anything, so happy to continue calling it keys for now

HenryMBaldwin: For now settled on calling it "Keys" for slightly higher accuracy and consistency

apps/lib/publisher_agent/config.go
HenryMBaldwin: Agreed, changed.

harryrackmil (approved):
Looks great!

PR #35, opened 2024-10-22
Aaron/sto 514 solana integration

akawalsky:
Going to work on getting this deployed now, but its worth looking over as is.

The idea here is that we will leverage our EVM signatures to write to the contract. Anyone can write to a feed account as long as the signature stored in the global config checks out. Only an admin (the deployer) should be able to update the config.

There are some tests, but probably are not exhaustive. Let me know if you can think of any other necessary tests. There are also some unit tests written in rust against the signature verification logic. We should add these to CI as well.


contracts/solana/stork/programs/stork/src/lib.rs
```diff
@@ -0,0 +1,244 @@
...
+            update_data.id,
+            update_data.temporal_numeric_value.timestamp_ns,
+            update_data.temporal_numeric_value.quantized_value,
+            update_data.publisher_merkle_root,
+            update_data.value_compute_alg_hash,
+            update_data.r,
+            update_data.s,
+            update_data.v,
+        ) {
+            return Err(StorkError::InvalidSignature.into());
```
HenryMBaldwin, line 55: Should probably switch to use the anchor err! macro. This should have the same result but be more idiomatic.

contracts/solana/stork/programs/stork/src/lib.rs
```diff
@@ -0,0 +1,244 @@
...
+        }
+
+        require_gte!(
+            update_data.temporal_numeric_value.timestamp_ns,
+            feed.latest_value.timestamp_ns,
+            StorkError::NoFreshUpdate
+        );
+
+        let amount_to_pay = config.single_update_fee_in_lamports;
```
HenryMBaldwin, line 64: amount_to_pay declare and init can be on the same line.

contracts/solana/stork/programs/stork/src/lib.rs
```diff
@@ -0,0 +1,244 @@
...
+        if payer.lamports()
+            < Rent::get()?
+                .minimum_balance(payer.data_len())
+                .saturating_add(amount_to_pay)
+        {
```
HenryMBaldwin, line 69: Not sure multiline is enhancing readability here, this can probably be all on the same line.

contracts/solana/stork/programs/stork/src/lib.rs
```diff
@@ -0,0 +1,244 @@
...
+            return err!(StorkError::InsufficientFunds);
+        };
+        let transfer_instruction =
+            system_instruction::transfer(payer.key, treasury.key, amount_to_pay);
```
HenryMBaldwin, line 73: Same here

contracts/solana/stork/programs/stork/src/lib.rs
```diff
@@ -0,0 +1,244 @@
...
+        anchor_lang::solana_program::program::invoke(
+            &transfer_instruction,
+            &[payer.to_account_info(), treasury.to_account_info()],
```
HenryMBaldwin, line 76: One line here too, or have the last argument of invoke() on a different line. Also since you're already importing anchor_lang::solana_program::system_instruction, it would probably make sense to just import program too in the same line (use  anchor_lang::solana_program::{system_instruction, program}) and cut out needing to use the path on program.

contracts/solana/stork/programs/stork/src/lib.rs
akawalsky: Seems like this is what rust's `fmt` command gives me ¯\_(ツ)_/¯

sdks/solana/stork-sdk/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+edition = "2021"
+
+
+[lib]
+crate-type = ["lib"]
+name = "stork_sdk"
+
+[dependencies]
+anchor-lang = "0.30.1"
+solana-program = ">=1.16.0"
```
HenryMBaldwin, line 16: solana-program dependency doesn't seem to be used in the sdk anywhere and can likely be removed.

HenryMBaldwin (approved):

PR #36, opened 2024-10-25
Add chainlink adapter

harryrackmil:
# Change
Adds a `stork_chainlink_adapter` directory which we can publish to npm to let people get data from the Stork EVM contract in their solidity contracts using the same interface they used for Chainlink.

Also adds a separate `examples/stork_chainlink_adapter` directory which is a hardhat project to demonstrate usage of this contract without polluting our evm contract or the stork_chainlink_adapter npm package.

# Testing
Deployed the Stork contract locally and used `interact.ts` to push some dummy data to it. Then deployed the `ExampleStorkChainlinkAdapter` contract locally pointed at that local Stork contract. I was able to get "latest round data" from the `ExampleStorkChainlinkAdapter` which had the price and timestamp I pushed to the underlying Stork contract.


contracts/evm/contracts/Stork.sol
```diff
@@ -67,6 +67,17 @@ abstract contract Stork is StorkGetters, StorkSetters, StorkVerify {
         return numericValue;
     }
 
+    function getTemporalNumericValueUnsafeV1(
```
harryrackmil, line 70: copied from `getTemporalNumericValueV1` but without the "stale value" check so we'll always return a latest value

contracts/stork_chainlink_adapter/contracts/IStorkTemporalNumericValueUnsafeGetter.sol
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity >=0.8.24 <0.9.0;
+
+import "./StorkStructs.sol";
+
+
+interface IStorkTemporalNumericValueUnsafeGetter {
```
harryrackmil, line 7: Rather than moving `Stork.sol` into our npm package, define an interface here which has the only function we need from `Stork.sol`.

contracts/stork_chainlink_adapter/contracts/StorkChainlinkAdapter.sol
```diff
@@ -0,0 +1,101 @@
...
+
+    function getTimestamp(uint256) external view returns (uint256) {
+        return latestTimestamp();
+    }
+
+    /*
+    * @notice This is exactly the same as `latestRoundData`, just including for parity with Chainlink
+    * Stork doesn't store roundId on chain so there's no way to access old data by round id
+    */
+    function getRoundData(
```
harryrackmil, line 57: other option here is to always fail if people try to call getRoundData since we won't be exposing old data, but that could cause problems with apps that are currently using this function on the chainlink aggregator.

contracts/stork_chainlink_adapter/contracts/StorkStructs.sol
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: Apache 2
+
+pragma solidity >=0.8.24 <0.9.0;
+
+contract StorkStructs {
+    struct TemporalNumericValue {
```
harryrackmil, line 6: copied from `StorkStructs.sol` in the evm package

contracts/stork_chainlink_adapter/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "@storknetwork/stork_chainlink_adapter",
```
harryrackmil, line 2: It might make sense to have a single npm package called `@storknetwork/sdk` or something.

akawalsky (commented):
Nice! A few initial comments and questions here.

contracts/stork_chainlink_adapter/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Integrate with Your Solidity Contracts
+1. Install the Stork Chainlink Adapter npm package in your project
+```
+npm i @storknetwork/stork_chainlink_adapter
+```
+
+2. Update your `remappings.txt` file to include:
+```
+@storknetwork/stork_chainlink_adapter/=node_modules/@storknetwork/stork_chainlink_adapter
```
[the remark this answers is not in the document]
harryrackmil: yea good point - seems to work without it

akawalsky (approved):

PR #37, opened 2024-10-28
Henry/sto 526 add solana support for chain pusher

HenryMBaldwin:
Added Solana implementation for chain pusher. ~~Currently "batch pushing" means just spinning up parallel single push go routines, and is not batched at the transaction level. This will likely change in the future as discussed.~~ Batch pushing uses the same batching window argument as used in evm, but inside the BatchPushToContract() call it breaks the price updates up into groups of 4 per batch (transaction), then sends those those batch transactions in parallel. The arguments to the Solana push CLI are slightly different than for EVM, but still straight forward.

The contract bindings have fuzz tests, and several of the functions in solana.go have tests a well.


apps/lib/chain_pusher/solana.go
```diff
@@ -0,0 +1,411 @@
...
+		}
+	}
+
+	return polledVals, nil
+}
+
+func (sci *SolanaContractInteracter) BatchPushToContract(priceUpdates map[InternalEncodedAssetId]AggregatedSignedPrice) error {
+	// Create a rate limiter: 40 requests per 10 seconds (4 per second)
+	// with a burst of 10 to stay well within the limits
+	limiter := rate.NewLimiter(rate.Every(250*time.Millisecond), 10)
```
akawalsky, line 196: Can this be really slow for lots of assets? Since BatchPushToContract happens on the main loop my concern is that the price update messages from the chain and the websocket might fill up.

apps/lib/chain_pusher/solana.go
```diff
@@ -0,0 +1,411 @@
...
+}
+
+func (sci *SolanaContractInteracter) pushLimitedBatchUpdateToContract(priceUpdates map[InternalEncodedAssetId]AggregatedSignedPrice) error {
+
+	if len(priceUpdates) > MAX_BATCH_SIZE {
+		sci.logger.Error().Msg("Batch size exceeds limit, skipping update")
+		return nil
+	}
+
+	configAccount, _, err := solana.FindProgramAddress(
```
akawalsky, line 256: I like that the feed address is defined in the init call. Thoughts on doing the same for config and treasuries to save time?

apps/lib/chain_pusher/solana.go
```diff
@@ -0,0 +1,411 @@
...
+			if key == sci.payer.PublicKey() {
+				return &sci.payer
+			}
+			return nil
+		})
+	if err != nil {
+		return fmt.Errorf("failed to sign transaction: %w", err)
+	}
+
+	sig, err := confirm.SendAndConfirmTransaction(context.Background(), sci.client, sci.wsClient, tx)
```
akawalsky, line 394: Is it slow to confirm the transaction? Maybe only confirm if given a certain configuration to go faster?

apps/lib/chain_pusher/solana.go
```diff
@@ -0,0 +1,411 @@
...
+func (sci *SolanaContractInteracter) ListenContractEvents(ch chan map[InternalEncodedAssetId]InternalStorkStructsTemporalNumericValue) {
+	for _, feedAccount := range sci.feedAccounts {
+		sub, err := sci.wsClient.AccountSubscribe(feedAccount, rpc.CommitmentFinalized)
+		if err != nil {
+			sci.logger.Error().Err(err).Str("account", feedAccount.String()).Msg("Failed to subscribe to feed account")
+			continue
+		}
+		defer sub.Unsubscribe()
+
+		go func(sub *ws.AccountSubscription, feedAccount solana.PublicKey) {
```
akawalsky, line 109: I generally like to avoid inlining functions unless its really convenient to do so. Does this one need to be inlined?

contracts/solana/Anchor.toml
```diff
@@ -5,7 +5,7 @@ resolution = true
 skip-lint = false
 
 [programs.localnet]
-stork = "2TSL7JwuTu9co7yUizwuh8EVdd4d96vDo9JykQCw8SHi"
+stork = "3ZZfMdMb4Ar1X5Utk39nEa3TKmZSyUeirkUjbbuqBX4V"
```
akawalsky, line 8: Can you revert this?

go.mod
```diff
@@ -15,36 +15,61 @@ require (
...
 	github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2 // indirect
 	github.com/shirou/gopsutil v3.21.11+incompatible // indirect
 	github.com/spf13/pflag v1.0.5 // indirect
+	github.com/streamingfast/logging v0.0.0-20230608130331-f22c91403091 // indirect
 	github.com/supranational/blst v0.3.11 // indirect
 	github.com/tklauser/go-sysconf v0.3.13 // indirect
 	github.com/tklauser/numcpus v0.7.0 // indirect
 	github.com/x448/float16 v0.8.4 // indirect
 	github.com/yusufpapurcu/wmi v1.2.4 // indirect
+	go.mongodb.org/mongo-driver v1.11.0 // indirect
```
akawalsky, line 63: Are these actually all necessary? Why should we need anything related to mongo?

apps/lib/chain_pusher/solana.go
```diff
@@ -0,0 +1,411 @@
...
+	// Create a rate limiter: 40 requests per 10 seconds (4 per second)
+	// with a burst of 10 to stay well within the limits
+	limiter := rate.NewLimiter(rate.Every(250*time.Millisecond), 10)
+
+	var wg sync.WaitGroup
+	errChan := make(chan error, len(priceUpdates))
+
+	priceUpdatesBatch := make(map[InternalEncodedAssetId]AggregatedSignedPrice)
+	i := 0
+	for encodedAssetId, priceUpdate := range priceUpdates {
```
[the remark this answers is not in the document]
akawalsky: Would also make this piece of it testable which would be nice

akawalsky (commented):
A few initial comments here

contracts/solana/Anchor.toml
akawalsky: yep. Theres a good chance we'll want to change this before we actually go live but lets leave it this way for now

go.mod
HenryMBaldwin: Most of these (including that mongo one) are dependencies for go-solana unfortunately.

apps/lib/chain_pusher/solana.go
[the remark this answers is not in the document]
HenryMBaldwin: I actually initially wrote batching and pushing as two separate loops, and then went and did it this way so that it'd be faster. I can break out batching into a separate function, I agree readability and test ability would improve, and I'd imagine that the performance hit would be negligible.

apps/lib/chain_pusher/solana.go
HenryMBaldwin: Yeah confirming is pretty slow, however it doesn't slow down the sending of the transaction. This isn't actually blocking other batches from going out from inside batchpushtocontract. I do see that batchpushtocontract blocks the select statement in Run() tho. I think I'd prefer spinning up a separate go routine for the batch push to contract call in the select statement in run. That wait we can confirm by default without actually slowing anything down. How do you feel about that?

apps/lib/chain_pusher/solana.go
HenryMBaldwin: The rate limiter would slow it down yes but this limit is imposed by the public solana rpc. As mentioned below I do think changing the batchpushtocontract call to a go routine is the right move, but this would also change the function signature because the rate limiter would need to be shared between any batchpushtocontract instance, which would mess with the evm pusher too. I don't necessarily think its the worst thing to have rate limit as a flag in both, especially because different rpcs on both evm and solana side will have different rate limits. How do you feel about this?

apps/lib/chain_pusher/solana.go
```diff
@@ -0,0 +1,427 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to sign transaction: %w", err)
+	}
+
+	sig, err := sci.client.SendTransaction(context.Background(), tx)
+	if err != nil {
+		return fmt.Errorf("failed to send transaction: %w", err)
+	}
+	// check for confirmation without blocking
+	go sci.confirmTransaction(sig)
```
akawalsky, line 406: This might kick off a ton of go routines. It would probably be better to throw this onto a channel and use a thread pool in a separate go routine to do these confirmations.

HenryMBaldwin: Changed. I feel very strongly that confirmation should never block updates, so I moved to a limited worker pool (10 go routines for now, but no strong opinions on this number) to limit cpu overhead while using an unbounded queue to prevent blocking, which a normal buffered channel might do. After the inital lag time for the first confirmations to hit, the (avg) confirmation rate then keeps pace with the submission rate in my testing, so actual unbounded growth is unlikely, however imo preferable to blocking. Let me know your thoughts.

HenryMBaldwin: Its worth noting that we can actually calculate the maximum number of necessary workers with:

$$
\text{workers} = \frac{txs}{block}
$$

expanded:

$$
\text{workers} = \frac{\lceil \frac{\text{assets}}{4} \rceil}{\text{batchwindow}} \times \text{blocktime}
$$ 

where:

$$
\begin{align}
&\text{assets} = \text{number of assets} \\
&\text{batchwindow} = \text{batching window in seconds} \\
&\text{blocktime} = \text{average solana blocktime in seconds/block}
\end{align}
$$

This of course assumes every asset gets an update with great enough change in every batching window, which from my testing so seems not to be the case. In any case, not sure how complicated we want to make it, or what reasonable upper bounds are for the number of assets one pusher will ever be running for, But I can implement this math at interactor initialization so at least the number of workers is not a bit arbitrary like it is now.

apps/lib/chain_pusher/solana.go
```diff
@@ -0,0 +1,524 @@
...
+	}
+
+	avgBlockTime, err := sci.getRecentAvgBlockTime()
+	if err != nil {
+		logger.Fatal().Err(err).Msg("Failed to get recent average block time")
+		return nil, err
+	}
+	numWorkers := int(math.Ceil(math.Ceil(float64(len(assetConfig.Assets))/float64(sci.batchSize)) * avgBlockTime.Seconds()))
+	//add 1 for safety
+	numWorkers += 1
```
harryrackmil, line 144: Can you walk me through the reasoning behind this formula? If these workers are just logging whether a transaction was confirmed or not, couldn't we just have a single worker working through the queue of signatures at whatever speed? or at least a fixed size pool. It makes me a little nervous that we could spin up a huge number of threads here.

apps/lib/chain_pusher/solana.go
```diff
@@ -0,0 +1,524 @@
...
+	wg := sync.WaitGroup{}
+	for _, feedAccount := range sci.feedAccounts {
+		wg.Add(1)
+		go sci.listenSingleContractEvent(ch, feedAccount, &wg)
+	}
+	// Wait indefinitely
+	wg.Wait()
+}
+
+func (sci *SolanaContractInteracter) listenSingleContractEvent(ch chan map[InternalEncodedAssetId]InternalStorkStructsTemporalNumericValue, feedAccount solana.PublicKey, wg *sync.WaitGroup) {
```
harryrackmil, line 239: why do we need these `Internal` types? are they different from the existing non-internal types?

apps/lib/chain_pusher/solana.go
```diff
@@ -0,0 +1,524 @@
...
+		return solana.Signature{}, fmt.Errorf("Batch size exceeds limit, skipping update")
+	}
+
+	randomId, err := rand.Int(rand.Reader, big.NewInt(256))
+	if err != nil {
+		return solana.Signature{}, fmt.Errorf("failed to generate random ID: %w", err)
+	}
+	treasuryId := uint8(randomId.Uint64())
+
+	treasuryAccount := sci.treasuryAccounts[treasuryId]
```
harryrackmil, line 395: this pattern of using a pool of treasury accounts and picking one randomly for each push is for security reasons?

apps/lib/chain_pusher/solana.go
```diff
@@ -0,0 +1,524 @@
...
+		feedAccount := sci.feedAccounts[encodedAssetId]
+
+		//convert the quantized price to bin.Int128
+		quantizedPriceBigInt := new(big.Int)
+		quantizedPriceBigInt.SetString(string(priceUpdate.StorkSignedPrice.QuantizedPrice), 10)
+
+		quantizedPrice := bin.Int128{
+			Lo: quantizedPriceBigInt.Uint64(),
+			Hi: quantizedPriceBigInt.Rsh(quantizedPriceBigInt, 64).Uint64(),
+		}
```
harryrackmil, line 413: is the logic in this for loop tested? it's easy to mess things up when we're doing things like byte conversions - maybe we could pull out a `convertPriceUpdateToTemporalNumericValueEvmInput` helper method so we can unit test this

harryrackmil (commented):
Looking great! Left some nit comments + questions

apps/docs/chain_pusher.md
```diff
@@ -20,13 +20,15 @@ assets:
 
 See `sample.asset-config.yaml` for an example.
 
-2. Create a `private-key.secret` file. This file should contain the private key of the user's wallet. This is needed to pay gas/transaction fees.
+2. **For EVM chains:** Create a `private-key.secret` file. This file should contain the private key of the user's wallet. This is needed to pay gas/transaction fees.
+**For Solana**: Create a `keypair.json` file, this file should contain the Solana wallet keypair.
```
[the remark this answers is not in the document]
HenryMBaldwin: I agree with this

apps/lib/chain_pusher/solana.go
HenryMBaldwin: Agree, will implement

apps/lib/chain_pusher/solana.go
HenryMBaldwin: These types were already present in the pusher and I patterned their use off of evm. My assumption is have a standard type that doesn't come from autogenerated bindings. The evm type seems to be the same but the solana type is slightly different. I'll defer to @akawalsky on this though.

apps/lib/chain_pusher/solana.go
HenryMBaldwin: see [here](https://github.com/Stork-Oracle/stork-external/pull/37#discussion_r1826026127): 

the math on what is probably a bad case scenario, lets say 40 assets, 500ms blocktimes, 1 second batching window, is still only 5 workers. A more realistic scenario, 10 assets, 400ms blocktimes, 5 second batching window, is just 1 worker. Assuming the prices of all assets are always being pushed (which depends on volatility and is a worst case), this calculates the minimum number of workers required for them to always be able to keep up with queue. The unbounded queue is still maintained as a buffer against impermanent slowdowns we cant control like blocktime slowdowns and slow rpc responses. It is possible that the rpc calls would be fast enough for one worker to churn through the queue sequentially and still keep up, however from my understanding go routines are so lightweight that its [practical for 10s to 100s of thousands to be running without major performance hit](https://go.dev/doc/faq#goroutines), so I figure setting the number of workers at sort of the upper bound for what is necessary to prevent blocking new updates or queue growth creep is fine, especially if its pretty much on the order of 10 or less in practical scenarios.

apps/lib/chain_pusher/contract_bindings/solana/accounts.go
```diff
@@ -0,0 +1,137 @@
+// Code generated by https://github.com/henrymbaldwin/solana-anchor-go. DO NOT EDIT.
```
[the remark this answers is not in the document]
HenryMBaldwin: Yes, aiming to put in the pr this week. Docs will changes if/when it gets merged.

apps/lib/chain_pusher/solana.go
akawalsky: Confirming Henry's explanation is correct

akawalsky (commented):
Couple more comments here

apps/lib/chain_pusher/solana.go
```diff
@@ -0,0 +1,533 @@
...
+		batchSize:           batchSize,
+		confirmationInChan:  confirmationInChan,
+	}
+
+	avgBlockTime, err := sci.getRecentAvgBlockTime()
+	if err != nil {
+		logger.Fatal().Err(err).Msg("Failed to get recent average block time")
+		return nil, err
+	}
+	numWorkers := int(math.Ceil(math.Ceil(float64(len(assetConfig.Assets))/float64(sci.batchSize)) * avgBlockTime.Seconds()))
```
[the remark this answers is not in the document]
HenryMBaldwin: Fair enough. 10 should be plenty even for worst case.

apps/lib/chain_pusher/solana.go
akawalsky: It is for write throughput reasons. If you have multiple clients pushing and they're all trying to interact with the same treasury account it could create write concurrency issues.

akawalsky (approved):

PR #46, opened 2024-11-12
[Stork Publisher Agent] Authenticate via Signature

harryrackmil:
# Change
Use the public key/private key pair that every publisher already has to authenticate with Stork rather than using auth tokens.

# Testing
- [X] Unit tests
- [X] ran the whole EVM system locally and authenticated successfully without auth token
- [x] deploy to dev


apps/lib/publisher_agent/config.go
```diff
@@ -13,7 +13,6 @@ import (
 )
 
 var Hex32Regex = regexp.MustCompile(`^0x[0-9a-fA-F]+$`)
-var StorkAuthRegex = regexp.MustCompile(`^[A-Za-z0-9+/]+={0,2}$`)
```
harryrackmil, line 16: remove StorkAuth from the publisher agent completely

apps/lib/signer/cgo_flags.go
```diff
@@ -1,7 +1,7 @@
 package signer
 
 /*
-#cgo LDFLAGS: -L./rust/stork/target/release -lstork
+#cgo LDFLAGS: -L./rust/stork/target -L./../../../../.lib -lstork
```
harryrackmil, line 4: needed to do this to run locally after building with `make rust` - could be doing something wrong though

apps/lib/signer/signer.go
```diff
@@ -20,6 +22,19 @@ import (
 	"github.com/rs/zerolog"
 )
 
+// gmorkworld in ascii
+const StorkMagicNumber = "103109111114107119111114108100"
```
harryrackmil, line 26: first pass at a magic number - converted "gmorkworld" to ascii and concatted together all the values.

apps/lib/signer/signer.go
```diff
@@ -20,6 +22,19 @@ import (
...
+const StorkAuthAssetId = "STORKAUTH"
+const StarkEncodedStorkAuthAssetId = "0x53544f524b41555448000000000000007361757468"
```
harryrackmil, line 28: hardcode the asset id computed from STORKAUTH with our hardcoded oracle id `sauth` so we don't need to recompute it every time.

apps/lib/signer/signer.go
```diff
@@ -122,10 +137,10 @@ func (s *StarkSigner) SignPublisherPrice(publishTimestamp int64, asset string, q
 	assetHex := hex.EncodeToString([]byte(asset))
 	assetHexPadded := assetHex
 	if len(assetHex) < 34 {
-		assetHexPadded = "0x" + assetHex + strings.Repeat("0", 32-len(assetHex))
+		assetHexPadded = assetHex + strings.Repeat("0", 32-len(assetHex))
```
harryrackmil, line 140: I think this was a potential bug before? we were adding `0x` only if the input string was length <34. adding it outside of the if now.

apps/lib/signer/signer.go
```diff
@@ -185,6 +200,92 @@ func (s *StarkSigner) GetSignatureType() SignatureType {
 	return StarkSignatureType
 }
 
+type StorkAuthSigner interface {
```
harryrackmil, line 203: this is a thin wrapper around a `Signer` object

apps/lib/signer/signer.go
```diff
@@ -185,6 +200,92 @@ func (s *StarkSigner) GetSignatureType() SignatureType {
...
+func NewEvmAuthSigner(privateKeyStr EvmPrivateKey, logger zerolog.Logger) (*EvmAuthSigner, error) {
+	evmSigner, err := NewEvmSigner(privateKeyStr, logger)
+	if err != nil {
+		return nil, err
+	}
+	return &EvmAuthSigner{evmSigner: evmSigner}, nil
+}
+
+func (s *EvmAuthSigner) SignAuth(publishTimestamp int64) (string, string, string, error) {
+	timestampedSignature, _, err := s.evmSigner.SignPublisherPrice(publishTimestamp, StorkAuthAssetId, StorkMagicNumber)
```
harryrackmil, line 221: calls the existing SignPublisherPrice function but with a fixed asset id and quantizedValue

apps/lib/signer/signer.go
```diff
@@ -185,6 +200,92 @@ func (s *StarkSigner) GetSignatureType() SignatureType {
...
+
+func (s *StarkAuthSigner) GetAuthHeaders() (http.Header, error) {
+	publicKey := s.starkSigner.GetPublisherKey()
+	signatureType := s.starkSigner.GetSignatureType()
+	timestamp := time.Now().UnixNano()
+	signatureR, signatureS, _, err := s.SignAuth(timestamp)
+	if err != nil {
+		return nil, fmt.Errorf("failed to sign auth header: %v", err)
+	}
+	header := http.Header{
```
akawalsky, line 279: Spoke offline about potentially changing this interface slightly

akawalsky (approved):
Looks good, lets make sure the behavior changes related to hex conversions still work

apps/docs/publisher_agent.md
```diff
@@ -17,19 +17,18 @@ Here we open a websocket at `ws://localhost:5216/publish` to receive price updat
 }
 ```
 
-Once you've gotten a StorkAuth key from Stork, you can use the [generate_keys.py](../../python/src/generate_keys/generate_keys.py) script to generate your EVM and/or Stork keys and to build the `keys.json` file. Make sure not to check this file into version control or share it in any way. 
+You can use the [generate_keys.py](../../python/src/generate_keys/generate_keys.py) script to generate your EVM and/or Stork keys and to build the `keys.json` file. Make sure not to check this file into version control or share it in any way.
```
harryrackmil, line 20: previous process was:
1. get an auth key from Stork
2. generate keypairs and configure
3. send your public key to Stork
4. run publisher agent

new process is:
1. generate keypairs and configure
2. send your public key to stork
3. run publisher agent

apps/docs/publisher_agent.md
```diff
@@ -17,19 +17,18 @@ Here we open a websocket at `ws://localhost:5216/publish` to receive price updat
...
-Once you've gotten a StorkAuth key from Stork, you can use the [generate_keys.py](../../python/src/generate_keys/generate_keys.py) script to generate your EVM and/or Stork keys and to build the `keys.json` file. Make sure not to check this file into version control or share it in any way. 
+You can use the [generate_keys.py](../../python/src/generate_keys/generate_keys.py) script to generate your EVM and/or Stork keys and to build the `keys.json` file. Make sure not to check this file into version control or share it in any way. 
 
 You can also generate your own keys and build your own `keys.json` file - it follows the structure of the [Keys struct](../lib/publisher_agent/config.go):
 ```json
 {
   "EvmPrivateKey": "0x8b558d5fc31eb64bb51d44b4b28658180e96764d5d5ac68e6d124f86f576d9de",
   "EvmPublicKey": "0x99e295e85cb07c16b7bb62a44df532a7f2620237",
-  "OracleId": "oracl",
-  "StorkAuth": "fake_auth"
```
harryrackmil, line 28: remove all references to StorkAuth from this readme since publishers will no longer use it

apps/docs/publisher_agent.md
```diff
@@ -39,17 +38,18 @@ The available environment variables are:
 | STORK_STARK_PRIVATE_KEY |
 | STORK_STARK_PUBLIC_KEY |
 | STORK_ORACLE_ID |
-| STORK_AUTH |
-| STORK_PULL_BASE_AUTH |
+| STORK_PULL_BASED_AUTH |
+
+**You'll need to send your Public keys to Stork before running your publisher agent so that we can whitelist them. NEVER SHARE YOUR PRIVATE KEYS WITH ANYONE, INCLUDING ANYONE CLAIMING TO BE A MEMBER OF STORK. WE WILL NEVER ASK YOU FOR IT.**
```
harryrackmil, line 43: moved this up from below - the agent will fail to hit the metadata endpoint and the registry until the key has been whitelisted, so it's better that publishers send us their public key before running the agent.

apps/lib/signer/cgo_flags.go
akawalsky: What are you doing to run locally? This shouldnt be necessary

akawalsky (approved):
Approved other than one cgo question

akawalsky (dismissed):

akawalsky (approved):

PR #59, opened 2024-11-28
Henry/sto 570 add tests to ci flow

HenryMBaldwin:


.github/workflows/ci.yml
```diff
@@ -183,3 +183,36 @@ jobs:
...
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            target/
+          key: ${{ runner.os }}-cargo-sui-${{ hashFiles('**/Cargo.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-cargo-sui-
+
+      - name: Install Sui
```
akawalsky, line 208: This looks like it took 33 minutes. Can this be cached?

akawalsky: If not, is there a faster way to install this? At 33 minutes unfortunately don't think it will be practical to include this in the build.

HenryMBaldwin: Caching worked. Job at around 1.5 minutes now

PR #63, opened 2024-12-05
Henry/sui integration

HenryMBaldwin:
Sui integration with 4 main pieces:

<ul>
<li> Sui contract
<li> Sui contract CI workflow
<li> Sui chain pusher
<li> Sui contract example
</ul>

As well as relevant docs in the form of READMEs.


contracts/sui/contracts/sources/stork.move
```diff
@@ -0,0 +1,172 @@
...
+    const EInsufficientFee: u64 = 1;
+    const EFeedNotFound: u64 = 2;
+
+    // === Entry Functions ===
+
+    entry fun init_stork(
+        // admin capability
+        _: &AdminCap,
+        // the address of the Stork program
+        stork_sui_public_key: address,
```
akawalsky, line 30: Does SUI use a different native signing method than EVM? The intention for Solana is that we may one day want to include a Solana public key to use for signing instead of using the EVM default signing strategy. I'm not sure if we would want to do the same for SUI. If not, might be best to remove this.

akawalsky, line 30: Is this meant to represent the current contract address? If so, I'd suggest a more descriptive name.

akawalsky: Why is this necessary to store on the contract?

contracts/sui/contracts/sources/stork.move
```diff
@@ -0,0 +1,172 @@
...
+            state_id,
+            stork_state_version
+        );
+    }
+
+    // === Public Functions ===
+
+    // updates the price feed for an asset 
+    // input data is an assets update data signed with Storks EVM public key
+    public fun update_single_temporal_numeric_value_evm(
```
akawalsky, line 56: Why did you want to keep this as well as the multiple update method?

HenryMBaldwin: Mostly to leave it available as an option for us in the future or people who want to write their own pushers. Sort of in the mindset that its written, it works, if this was say a rest api endpoint it would feel natural to include it even if we don't personally use it in our clients, and because upgrading is a pain I'd rather not have to add it back in later. If you want to take it out we can, as we're not using it, however it feels natural to keep it in imo.

contracts/sui/contracts/sources/state.move
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    public fun get_total_fees_in_mist(stork_state: &StorkState, num_updates: u64): u64 {
+        assert!(stork_state.version == VERSION, EIncorrectVersion);
+        stork_state.get_single_update_fee_in_mist() * num_updates
+    }
+
+    // === Admin Functions ===
+
+    public fun update_single_update_fee_in_mist(
```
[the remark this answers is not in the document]
HenryMBaldwin: An object can only be mutated from within the module that defines it, so we could add wrappers in the stork::stork module like we did for updating, however unlike those update functions, which are more complicated than simple setters and getters and actually do other logic, these would essentially just be duplicate setters and getters that call these functions anyways.

akawalsky (approved):
Lets do a regular merge on this one

PR #67, opened 2024-12-28
update Plume explorer URLs

eyqs:


akawalsky (approved):
Thanks for the contribution @eyqs 

The Stork contract is actually already deployed to Plume mainnet, see here - https://phoenix-explorer.plumenetwork.xyz/address/0xacC0a0cF13571d30B4b8637996F5D6D774d4fd62

The reason this isnt in code is because we recently added the ability internally to deploy contracts programatically. When we are ready, we will deprecate this hardhat config.

Going to approve and merge this for now though.

PR #68, opened 2025-01-07
Send config to publisher metadata

harryrackmil:
# Change
* Remove all secrets from the StorkPublisherAgentConfig and move those secrets into a separate StorkPublisherAgentSecrets struct
* pass the config to the publisher metadata endpoint.

# Testing
[X] Ran locally and confirmed the json was written (without any secrets) and properly structured
[X] Running for two dev publishers and data is showing up without any secrets and properly structured


apps/lib/publisher_agent/metadata.go
```diff
@@ -104,9 +109,27 @@ func (p *PublisherMetadataReporter) getMetadata() PublisherMetadata {
...
+		ConfigJson:            config,
 	}
 }
 
+func getRedactedConfigJson(config StorkPublisherAgentConfig) string {
+	// copy the config so we're not mutating the underlying
+	configCopy := config
+
+	// redact
+	configCopy.StarkPrivateKey = ""
```
harryrackmil, line 121: a little messy - it would be cleaner to fully specify a redacted config object and explicitly convert the existing config into the redacted version. That seemed like it would be painful to maintain since we may add new fields to the config and forget to update here. Choosing instead to "blacklist" specific fields that we know are secret, since most configuration parameters won't be secret.

Right now these just show up as blanks in the persisted config - maybe worth setting them to the string "REDACTED" or something to make it clearer.

apps/lib/publisher_agent/metadata_test.go
```diff
@@ -0,0 +1,43 @@
...
+		EvmPrivateKey:   myEvmPrivateKey,
+		EvmPublicKey:    myEvmPublicKey,
+		StarkPrivateKey: myStarkPrivateKey,
+		StarkPublicKey:  myStarkPublicKey,
+		PullBasedAuth:   myPullBasedAuth,
+	}
+
+	redactedConfigJson := getRedactedConfigJson(config)
+
+	// original config isn't mutatated
```
harryrackmil, line 28: confirming that the original config isn't mutated by the redaction process

akawalsky (commented):
Maybe instead it would make more sense to split out the secrets to a separate config? That way they're never combined and we dont need to worry about forgetting to blacklist something in the future

apps/lib/publisher_agent/command.go
```diff
@@ -34,7 +34,7 @@ func runPublisherAgent(cmd *cobra.Command, args []string) error {
 	configFilePath, _ := cmd.Flags().GetString(ConfigFilePathFlag)
 	keysFilePath, _ := cmd.Flags().GetString(KeysFilePathFlag)
 
-	config, err := LoadConfig(configFilePath, keysFilePath)
+	config, secrets, err := LoadConfig(configFilePath, keysFilePath)
```
harryrackmil, line 37: turns out we only need to use the secrets inside this run command

apps/lib/publisher_agent/config.go
```diff
@@ -96,16 +96,16 @@ func readFile(path string) ([]byte, error) {
 	return data, nil
 }
 
-func LoadConfig(configFilePath string, keysFilePath string) (*StorkPublisherAgentConfig, error) {
+func LoadConfig(configFilePath string, keysFilePath string) (*StorkPublisherAgentConfig, *StorkPublisherAgentSecrets, error) {
 	configFileData, err := readFile(configFilePath)
 	if err != nil {
-		return nil, fmt.Errorf("failed to read config file: %w", err)
+		return nil, nil, fmt.Errorf("failed to read config file: %w", err)
```
harryrackmil, line 102: account for the extra returned struct in all error handling

apps/lib/publisher_agent/config.go
```diff
@@ -266,22 +264,43 @@ func LoadConfig(configFilePath string, keysFilePath string) (*StorkPublisherAgen
...
+	secrets := NewStorkPublisherAgentSecrets(
+		keys.EvmPrivateKey,
+		keys.StarkPrivateKey,
+		keys.PullBasedAuth,
+	)
+
+	return config, secrets, nil
+}
+
+type StorkPublisherAgentSecrets struct {
```
harryrackmil, line 283: remove all private keys from StorkPublisherAgentConfig and move them to this new struct

akawalsky (approved):

PR #69, opened 2025-01-10
Add data provider

harryrackmil:
# Change
Add a framework for reporting data from arbitrary sources to the publisher agent

# Testing
Ran this and a publisher agent together via docker-compose and confirmed prices looked reasonable and the publisher agent was able to parse and sign them


apps/lib/data_provider/command.go
```diff
@@ -0,0 +1,50 @@
...
+	RunE:  runDataProvider,
+}
+
+// required
+const ConfigFilePathFlag = "config-file-path"
+const WebsocketUrl = "ws-url"
+
+func init() {
+	DataProviderCmd.Flags().StringP(ConfigFilePathFlag, "c", "", "the path of your config json file")
+	DataProviderCmd.Flags().StringP(WebsocketUrl, "w", "", "the websocket url to write updates to")
```
harryrackmil, line 24: This could be configured in the config json if we wanted, but it felt like more of a run time configuration (I might want to test without a websocket url at first to just look at the prices)

apps/lib/data_provider/data_source.go
```diff
@@ -0,0 +1,31 @@
+package data_provider
+
+type dataSource interface {
+	// Add all value updates to updatesCh
+	Run(updatesCh chan DataSourceUpdateMap)
+	GetDataSourceId() DataSourceId
+}
+
+func buildDataSources(config DataProviderConfig) []dataSource {
+	// group by data source id to support batched feeds
```
harryrackmil, line 10: overkill for our current sources (right now every dataSource is 1:1 with a value) but we might want to support batching of feeds in the future

apps/lib/data_provider/data_source_registry.go
```diff
@@ -0,0 +1,12 @@
+package data_provider
+
+func GetDataSourceBuilder(dataSourceId DataSourceId) func([]DataProviderSourceConfig) []dataSource {
+	switch dataSourceId {
+	case UniswapV2DataSourceId:
```
harryrackmil, line 5: This switch statement is the only shared file that needs to change when we write a new integration

apps/lib/data_provider/random.go
harryrackmil, line 1: a simpler example than uniswap. Still using the scheduled data source concept

apps/lib/data_provider/runner.go
```diff
@@ -0,0 +1,26 @@
...
+	}
+}
+
+func (r *DataProviderRunner) Run() {
+	r.dataSources = buildDataSources(r.config)
+	for _, dataSource := range r.dataSources {
+		go dataSource.Run(r.updatesCh)
+	}
+
+	r.writer.Run(r.updatesCh)
```
harryrackmil, line 25: kick off all the data sources in goroutines and kick off a writer thread

apps/lib/data_provider/scheduled_data_source.go
harryrackmil, line 1: a helper to allow pulling on a regular cadence

apps/lib/data_provider/writer.go
```diff
@@ -0,0 +1,80 @@
...
+		}
+	}
+
+	for updateMap := range updateCh {
+		valueUpdates := make([]ValueUpdate, 0)
+		for _, update := range updateMap {
+			valueUpdate := ValueUpdate{
+				PublishTimestamp: update.Timestamp.UnixNano(),
+				ValueId:          update.ValueId,
+				Value:            fmt.Sprintf(`%.18f`, update.Value),
```
harryrackmil, line 54: convert to string so we don't report very small prices using scientific notation

apps/lib/data_provider/writer.go
```diff
@@ -0,0 +1,80 @@
...
+			Data: valueUpdates,
+		}
+		wsMessageBytes, err := json.Marshal(valueUpdateWebsocketMessage)
+		if err != nil {
+			w.logger.Error().Msgf("Error marshalling update %v: %v", valueUpdateWebsocketMessage, err)
+		}
+
+		w.logger.Debug().Msgf("Update: %s", string(wsMessageBytes))
+
+		if conn != nil {
```
harryrackmil, line 70: a user may not configure any websocket url when developing - they can just run the app in verbose mode to see the updates logged

akawalsky:
Im not seeing an entry point and the docker compose isnt working locally. How have you been running this?

apps/lib/data_provider/sources/random/factory.go
```diff
@@ -0,0 +1,31 @@
...
+var resourcesFS embed.FS
+
+type randomDataSourceFactory struct{}
+
+func (f *randomDataSourceFactory) Build(sourceConfig types.DataProviderSourceConfig) types.DataSource {
+	return newRandomDataSource(sourceConfig)
+}
+
+func (f *randomDataSourceFactory) GetSchema() (*gojsonschema.Schema, error) {
+	return utils.LoadSchema("resources/config_schema.json", resourcesFS)
```
harryrackmil, line 24: GetSchema is really a function of the data source class, not a specific instance of the data source, so I'm making it a function on the factory object. Might be worth renaming the factory interface to `DataSourceType` or `DataSourceClass` or something since it's doing more than just building data sources now.

apps/lib/data_provider/sources/scheduler.go
```diff
@@ -0,0 +1,47 @@
...
+type Scheduler struct {
+	updateFrequency time.Duration
+	getUpdate       func() (types.DataSourceUpdateMap, error)
+	handleErr       func(error)
+}
+
+func NewScheduler(
+	updateFrequency time.Duration,
+	getUpdate func() (types.DataSourceUpdateMap, error),
+	handleErr func(error),
```
harryrackmil, line 19: rather than passing a logger object to the scheduler, we pass handleErr function. In practice I expect most sources will call this with the `GetErrorLogHandler()` function provided in this file which just logs at a configurable level. Might make more sense to just pass a logger to the scheduler instead.

apps/lib/data_provider/sources/uniswap_v2/config.go
```diff
@@ -0,0 +1,12 @@
+package uniswap_v2
```
harryrackmil, line 1: this file is just the config object, in case we wanted to autogenerate from json schema (or generate json schema from this)

apps/lib/data_provider/sources/uniswap_v2/data_source.go
```diff
@@ -0,0 +1,153 @@
+package uniswap_v2
```
harryrackmil, line 1: the `data_source` file has minimal boilerplate - it's just the DataSource object, which only needs to implement `RunDataSource`, and a constructor for the DataSource object.

apps/lib/data_provider/types/model.go
```diff
@@ -0,0 +1,51 @@
+package types
```
harryrackmil, line 1: making `types` its own package so every other package is able to depend on it

apps/lib/data_provider/sources/ethereum_utils.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/ethereum/go-ethereum/ethclient"
+	"github.com/rs/zerolog"
+)
+
+const (
+	maxQueryAttempts = 3
+	baseRetryDelay   = 500 * time.Millisecond
+)
+
+func GetEthereumContract(
```
harryrackmil, line 22: pulled some of the reusable eth contract stuff out of the uniswap integrations since it's a lot of code and will probably be used in other integrations - might make sense to put it in the `utils` directory. Leaving in `sources` for now because it will only be used by data sources

apps/scripts/generate_data_provider_import_file.py
```diff
@@ -0,0 +1,25 @@
+ import os
```
harryrackmil, line 1: Generate a file which will explicitly force importing each individual data source package. Using python since it was a little less verbose than the go version of this script, but happy to convert to go if we'd like.

We sort the package names here so it should generate the same file consistently. We can add a github action which runs this script and checks to see if the result of this script is different than the currently committed imports file, and block merging if they disagree

akawalsky: Is there a world where this also adds to the json schema config at the top level to support oneOf all of the resources included using a schema pointer? If so, we could eliminate the awkwardness of having to load each config individually.

apps/lib/data_provider/writer.go
```diff
@@ -0,0 +1,82 @@
...
+	for {
+		err := w.runWriteLoop(updateCh)
+		if err != nil {
+			w.logger.Info().Err(err).Str("url", w.wsUrl).Msgf("Loop exited - waiting %s to resume", reconnectDuration)
+			time.Sleep(reconnectDuration)
+		}
+	}
+}
+
+func (w *WebsocketWriter) runWriteLoop(updateCh chan types.DataSourceUpdateMap) error {
```
akawalsky, line 40: We should prob build in reconnect logic for the case where the publisher agent is restarted

apps/lib/data_provider/sources/uniswap_v2/config.go
akawalsky: Would it also be worth having a config_test to confirm that the config is deserialized as expected?

apps/lib/data_provider/command.go
akawalsky: What do you think of making this a bit more generic - something like `-o` where the output could eventually take the form of 
1. ws interface
2. http interface
3. a file buffer interface

which could be denoted by ws(s)://, http(s)://, file://, s3:// etc

akawalsky (commented):
This is looking awesome, just a few more comments/questions

.github/workflows/ci.yml
```diff
@@ -49,6 +49,29 @@ jobs:
         env:
           TARGETPLATFORM: linux/amd64
 
+  data-provider-codegen-check:
```
harryrackmil, line 52: run the codegen script and fail if there's any git diff. Tested this by making a small change in one of the individual config_schema.json, not running the codegen script, and confirming the github action fails.

apps/lib/data_provider/resources/config_schema.json
```diff
@@ -0,0 +1,138 @@
+{
```
harryrackmil, line 1: auto-generated by the `update_shared_data_provider_code.py` script

apps/lib/data_provider/resources/config_schema.json.template
```diff
@@ -0,0 +1,38 @@
+{
```
harryrackmil, line 1: template file for top level config_schema generation

apps/lib/data_provider/sources/random/factory.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"embed"
+
+	"github.com/Stork-Oracle/stork-external/apps/lib/data_provider/sources"
+	"github.com/Stork-Oracle/stork-external/apps/lib/data_provider/types"
+	"github.com/Stork-Oracle/stork-external/apps/lib/data_provider/utils"
+)
+
+var RandomDataSourceId types.DataSourceId = types.DataSourceId(utils.GetCurrentDirName())
```
harryrackmil, line 11: data source id should be the same as the directory name (makes config schema generation easier)

apps/lib/data_provider/writer.go
harryrackmil: This is handled already I believe - if we ever get an error when writing we'll stop the loop and return an error. This function is called inside a `for` loop in the `Run` function above so it should wait 5 seconds and then reconnect.

apps/scripts/update_shared_data_provider_code.py
```diff
@@ -0,0 +1,80 @@
...
+        for dir in directories:
+            f.write(f'import _ "github.com/Stork-Oracle/stork-external/apps/lib/data_provider/sources/{dir}"\n')
+    print(f"Go imports file generated at {IMPORT_SOURCES_FILE}")
+
+
+def generate_config_schema(directories):
+    source_config_schemas = []
+    data_source_ids = []
+
+    conditional_source_config_schema_template_str = """
```
harryrackmil, line 29: we'll use this `if/then` to make sure that we're checking the appropriate schema for this data source - if we just check that the config matches any defined schema we'll wind up being as permissive as the most permissive schema.

apps/scripts/update_shared_data_provider_code.py
```diff
@@ -0,0 +1,80 @@
...
+
+    data_source_ids_str = ",".join([f'"{x}"' for x in data_source_ids])
+
+    template_data = {
+        "source_config_schemas": source_config_schema_str,
+        "data_source_ids": data_source_ids_str
+    }
+
+    with open(CONFIG_SCHEMA_TEMPLATE_FILE, 'r') as template_file:
+        template_json = template_file.read()
```
harryrackmil, line 61: after wrapping each config schema in an if/else and concatenating them all together, fill in the `config_schema.json.template` file

apps/lib/data_provider/sources/random/config_test.go
```diff
@@ -0,0 +1,95 @@
...
+	_, err := utils.LoadSchema(resourcesFS)
+	assert.NoError(t, err)
+}
+
+func TestValidConfig(t *testing.T) {
+	schema, err := utils.LoadSchema(resourcesFS)
+	assert.NoError(t, err)
+
+	configStr := `
+	{
```
harryrackmil, line 21: note that we're not testing the top level config, we're specifically testing the json within the `config` tag, applying only to this source

harryrackmil:
> Im not seeing an entry point and the docker compose isnt working locally. How have you been running this?

I committed the changes I'd made locally to the docker-compose so now you can run it from docker-compose:
```
docker-compose up --build data-provider
```

apps/scripts/generate_data_provider_import_file.py
harryrackmil: The schema pointer `$ref` concept doesn't work so nicely since the config schemas are spread across different `embed.FS` instances in different packages - we'd need to write some custom schema loader code to use pointers, meaning the schema file couldn't really be used directly since people would need to use our custom loader. 

Instead, I'm using the codegen script to generate a big flat config schema file which contains all the individual config schemas. We would have needed to do some codegen for the pointer approach anyway in order to update each pointer when we add new sources - this way we have a single interpretable config schema which doesn't require any custom loading code.

apps/lib/data_provider/command.go
harryrackmil: makes sense - added a `Writer` interface and a `GetWriter` function where we can do some branching in the future. Currently it will fail if the `--output-address` isn't either blank or prefixed with `ws://`

apps/docs/data_provider.md
```diff
@@ -0,0 +1,52 @@
+# Data Provider
+The Stork Data Provider is a framework to pull arbitrary numeric data across many sources. It can be used on its own, or run alongside the Stork Publisher Agent to sign the data and send it to the Stork Network.
+
+## Adding a New Data Source
+If you want to report data from a data source which does not already have an [integration](../lib/data_provider/sources), you can add your own.
+
+To add a new source:
+1. Add a [package](../lib/data_provider/sources/random) in the [sources directory](../lib/data_provider/sources) with your data source's name
+1. Run `python3 ./apps/scripts/update_shared_data_provider_code/py` to generate some framework code so that the framework is aware of your new source.
```
akawalsky, line 9: Numbers here seem to have gotten messed up somehow

akawalsky: Oh interesting. Is that a common practice? Not sure I've seen that before

apps/lib/data_provider/sources/raydiumclmm/data_source.go
```diff
@@ -0,0 +1,178 @@
...
+	apiKey := ""
+	if len(raydiumCLMMConfig.ProviderApiKeyEnvVar) > 0 {
+		var exists bool
+		apiKey, exists = os.LookupEnv(raydiumCLMMConfig.ProviderApiKeyEnvVar)
+		if !exists {
+			panic("env var with name " + raydiumCLMMConfig.ProviderApiKeyEnvVar + " is not set")
+		}
+	}
+
+	rpcClient := rpc.New(raydiumCLMMConfig.HttpProviderUrl + apiKey)
```
akawalsky, line 108: I think the api key should just be encoded directly in the URL here

ACK101101 (approved):

akawalsky (approved):

PR #70, opened 2025-01-10
Henry/sto 617 aptos contract implementation

HenryMBaldwin:
Aptos Contract, Admin CLI, and CI workflow


contracts/aptos/contracts/sources/state.move
```diff
@@ -0,0 +1,195 @@
...
+    ): StorkState {
+        StorkState {
+            stork_evm_public_key,
+            single_update_fee_in_octas,
+            owner,
+        }
+    }
+
+    /// Moves a StorkState to the given signer
+    package fun move_state(self: StorkState, signer: &signer) {
```
akawalsky, line 40: What prevents anyone other than an admin from calling these funcs?

akawalsky (commented):
As far as I can tell this looks good. Just one comment but we will be sure to get this fully audited.

Also, what is the upgrade pattern here?

contracts/aptos/contracts/sources/stork.move
```diff
@@ -0,0 +1,597 @@
...
+    /// Updates a single temporal numeric value using EVM signature for verification
+    public entry fun update_single_temporal_numeric_value_evm(
+        // signer of the transaction to pay the fee
+        signer: &signer,
+        // asset id
+        asset_id: vector<u8>,
+        // temporal numeric value timestamp ns
+        temporal_numeric_value_timestamp_ns: u64,
+        // temporal numeric value magnitude
+        temporal_numeric_value_magnitude: u128,
```
akawalsky, line 72: Why do we call this magnitude? Is it because we cannot encode negative numbers directly?

HenryMBaldwin: Yeah its has to be made into our custom I128 on-chain. move only supports unsigned ints natively.

PR #74, opened 2025-01-15
Henry/sto 641 fix stork external ci

HenryMBaldwin:
I've removed the ignore main for running this workflow on push. 

As it turns out, branches are restricted on which other branches caches they can use, however every branch can use the default branches cache, in this case main. This means, if we allow main to run the CI, it will generate caches, which then all branches will be able to pull from.

I have confirmed this in a private test repository.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,8 +2,6 @@ name: CI
 
 on:
   push:
-    branches:
-      - main
 
 jobs:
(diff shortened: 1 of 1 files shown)


HenryMBaldwin:
I canceled the workflows to save on unnecessary usage, which should be fixed after this is merged lol.

akawalsky (approved):

PR #75, opened 2025-01-16
Data Provider Skeleton Mockup

harryrackmil:
We'd like to make adding a new Data Provider source easier by auto-generating some skeleton code which the user can then fill in so they don't have to think too much about the framework.

In this PR I've manually written some suggested skeleton code for a new data source called `skeleton`, which can hopefully serve as a starting model for our templating

I think the ideal flow for a user would be a CLI-like tool where we ask them to give us the name of their dataSource in `TitleCase`. We would then automatically
1. validate the string they gave us was valid (no underscores/spaces etc, not using the name of an existing dataSource)
2. do some templating to auto-generate all the files in this PR (including the TODO comments) with their dataSource name in place of`skeleton`. We can figure out `camelCase`,`snake_case` and `flatcase` versions of the dataSource name automatically and insert the appropriate one into the template.
3. run `python3 ./apps/scripts/update_shared_data_provider_code.py` to update a couple of shared framework files
4. Print a message to the user telling them to fill in all the TODOs (and maybe listing all the files containing a TODO)

Then the user just fills in the TODOs, gets their tests passing and submits the PR.

### DIFF
--- a/apps/lib/data_provider/configs/resources/source_config_schemas/skeleton.json
+++ b/apps/lib/data_provider/configs/resources/source_config_schemas/skeleton.json
@@ -8,6 +8,6 @@
     }
    // TODO: add all other possible fields from your config object
   },
-  "required": ["dataSourceId"], // TODO: add all other required fields here
+  "required": ["dataSource"], // TODO: add all other required fields here
   "additionalProperties": false
 }
--- a/apps/lib/data_provider/configs/resources/source_config_schemas/skeleton.json
+++ b/apps/lib/data_provider/configs/resources/source_config_schemas/skeleton.json
@@ -1,0 +1,13 @@
+{
+  "$id": "/resources/source_config_schemas/skeleton",
+  "type": "object",
+  "properties": {
+    "dataSource": {
+      "type": "string",
+      "const": "skeleton"
+    }
+   // TODO: add all other possible fields from your config object
+  },
+  "required": ["dataSourceId"], // TODO: add all other required fields here
+  "additionalProperties": false
+}
--- a/apps/lib/data_provider/configs/resources/source_config_schemas/skeleton.json
+++ b/apps/lib/data_provider/configs/resources/source_config_schemas/skeleton.json
@@ -6,8 +6,7 @@
       "type": "string",
       "const": "skeleton"
     }
-   // TODO: add all other possible fields from your config object
   },
-  "required": ["dataSource"], // TODO: add all other required fields here
+  "required": ["dataSource"],
   "additionalProperties": false
 }
--- a/apps/lib/data_provider/configs/source_config_tests/skeleton_test.go
+++ b/apps/lib/data_provider/configs/source_config_tests/skeleton_test.go
@@ -27,7 +27,8 @@ func TestValidSkeletonConfig(t *testing.T) {
 
 	sourceSpecificConfig, err := skeleton.GetSourceSpecificConfig(sourceConfig)
 	assert.NoError(t, err)
+	assert.NotNil(t, sourceSpecificConfig)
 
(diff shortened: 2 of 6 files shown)


apps/lib/data_provider/configs/source_config_tests/skeleton_test.go
```diff
@@ -0,0 +1,33 @@
...
+	assert.Equal(t, 1, len(config.Sources))
+
+	sourceConfig := config.Sources[0]
+
+	dataSourceId, err := utils.GetDataSourceId(sourceConfig.Config)
+	assert.NoError(t, err)
+	assert.Equal(t, skeleton.SkeletonDataSourceId, dataSourceId)
+
+	sourceSpecificConfig, err := skeleton.GetSourceSpecificConfig(sourceConfig)
+	assert.NoError(t, err)
```
harryrackmil, line 29: generate tests to make sure the config is valid and can be loaded, that we can extract the data source id from it, and that we can extract the SkeletonConfig.

The user just has to fill in a valid config string and write some asserts

apps/lib/data_provider/sources/skeleton/config.go
```diff
@@ -0,0 +1,8 @@
+package skeleton
+
+import "github.com/Stork-Oracle/stork-external/apps/lib/data_provider/types"
+
+type SkeletonConfig struct {
+	DataSource types.DataSourceId `json:"dataSource"` // required for all Data Provider Sources
```
harryrackmil, line 6: All configs need to have a DataSource field

The user just needs to add configuration details specific to their dataSource

apps/lib/data_provider/sources/skeleton/data_source_test.go
```diff
@@ -0,0 +1,9 @@
+package skeleton
+
+import (
+	"testing"
+)
+
+func TestSkeletonDataSource(t *testing.T) {
+	// TODO: write some unit tests for your data source
```
harryrackmil, line 8: very open ended - didn't want to auto-generate unit tests that actually run the integration in case the integration hits external services

apps/lib/data_provider/sources/skeleton/init.go
```diff
@@ -0,0 +1,30 @@
+package skeleton
```
harryrackmil, line 1: this file is fully auto-generated - user doesn't need to touch it

ACK101101: maybe we can add a comment at the top of the file saying that it is autogenerated and to please not touch - similar to mockery

ACK101101 (commented):
i'm not 100% sure what i should be looking for but it seems like a good skeleton. should we provide an example implementation to make it even clearer?
also, what do you think about providing an explicit function in the flow for transformations? too opinionated?

apps/lib/data_provider/sources/skeleton/init.go
harryrackmil: hmm I'm not sure if we want to actually stop people from changing this file if they want to - it wouldn't be crazy for some advanced user to make changes to the factory code.

apps/lib/data_provider/configs/resources/source_config_schemas/skeleton.json
```diff
@@ -0,0 +1,12 @@
+{
```
harryrackmil, line 1: removed the `//TODO` comments from here because it's invalid json and was causing some framework tests to fail. We should probably explicitly say in the docs that users will need to edit this file once they write their config object. 

If they try to configure anything other than data source id with this default schema, validation will fail. So hopefully that will force users to update this file even without the TODOs

PR #78, opened 2025-01-22
Alexander/sto 646 codegen for data provider

ACK101101:


apps/lib/data_provider/gen.go
```diff
@@ -0,0 +1,346 @@
...
+
+	return nil
+}
+
+func validateDataProviderName(dataProviderName string, basePath string) error {
+	if !validatePascalCase(dataProviderName) {
+		return fmt.Errorf("data provider name must be in PascalCase. Please try again.")
+	}
+
+	dataSourcesDir := basePath + "/apps/lib/data_provider/sources"
```
harryrackmil, line 59: let's move this logic out into a `checkForDuplicateDataSource` function or something since it's a nontrivial amount of code. Makes it easier to read the `validateDataProviderName` function and understand the requirements at a high level

apps/lib/data_provider/gen.go
```diff
@@ -0,0 +1,346 @@
...
+	}
+
+	inputData := struct {
+		PascalStr string
+		LowerStr  string
+		CamelStr  string
+	}{
+		PascalStr: pascalName,
+		LowerStr:  pascalToLower(pascalName),
+		CamelStr:  pascalToCamel(pascalName),
```
harryrackmil, line 321: nit: rather than doing these case conversions once per file maybe we can do it once at the start and then pass them through to each generate command? doesn't make a huge difference - just nice to keep the `generate` command as low-business-logic as possible so we don't have to touch it often

apps/lib/data_provider/command.go
```diff
@@ -10,21 +12,46 @@ import (
 	"github.com/spf13/cobra"
 )
 
-var DataProviderCmd = &cobra.Command{
```
ACK101101, line 13: renamed DataProviderCmd to StartDataProviderCmd

apps/lib/data_provider/configs/resources/templates/shared/import_sources.gotmpl
```diff
@@ -0,0 +1,6 @@
+// @path: apps/lib/data_provider/configs/import_sources.go
+package configs
+
+// Autogenerated by data_provider generator
+{{range .PackageNames}}import _ "github.com/Stork-Oracle/stork-external/apps/lib/data_provider/sources/{{.}}"
```
ACK101101, line 5: same as above, but isn't as ugly imo

apps/lib/data_provider/configs/resources/data_provider_config.schema.json
```diff
@@ -18,24 +18,15 @@
...
     }
   },
-  "required": [
-    "sources"
-  ],
+  "required": ["sources"],
   "additionalProperties": false
-}
\ No newline at end of file
+}
```
ACK101101, line 32: ^ a bit uglier, left a comment about this on data_provider_config.schema.gotmpl

Makefile
```diff
@@ -14,3 +14,17 @@ help:
 		sub(/:.*/, "", $$1); \
 		printf "  %-30s %s\n", $$1, helpMessage; \
 	}' $(MAKEFILE_LIST)
+
+## Install the stork-data-provider binary
+.PHONY: install-data-provider-cli
+install-data-provider-cli:
+	@echo "Installing stork-data-provider..."
+	@go build -o $(shell go env GOPATH)/bin/stork-data-provider ./apps/cmd/data_provider
+	@echo "Successfully installed stork-data-provider. Run 'stork-data-provider help' to get started."
```
harryrackmil, line 23: this didn't work for me out of the box since `~/go/bin` wasn't in my $PATH - maybe we can also echo the installation location so users can run the binary directly even if their PATH is wrong

apps/lib/data_provider/configs/resources/data_provider_config.schema.json
harryrackmil: ah I don't mind this at all.

Also, couldn't we just update the template to look like the pretty-printed json? At least for things like this `"required": ["sources"],` which aren't autogenerated. Pretty sure I was just doing the automatic json-formatting because the output of the python script was much uglier

apps/lib/data_provider/configs/resources/templates/shared/data_provider_config.schema.json.gotmpl
```diff
@@ -18,7 +19,10 @@
             "description": "An arbitrary configuration object for a data source",
             "type": "object",
             "oneOf": [
-                $source_config_schemas
+              {{- range $i, $ref := .PackageNames }}
+                {{- if $i }},{{ end }}
+                {"$ref": "/resources/source_config_schemas/{{ $ref }}"}
+              {{- end }}
             ]
```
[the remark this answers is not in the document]
harryrackmil: I don't really mind this. Probably nicer than pre-building the block of sources in go code since we'd have to think about indentation. Let's maybe name the variable something other than $ref though since we're also using $ref as a literal value in this line

apps/lib/data_provider/configs/resources/templates/skeleton/config.gotmpl
harryrackmil, line 1: let's name this directory `templates/source` or something rather than `templates/skeleton`

apps/lib/data_provider/configs/resources/templates/skeleton/config.gotmpl
```diff
@@ -0,0 +1,11 @@
+// @path: /apps/lib/data_provider/sources/{{ .LowerStr }}/config.go
+// Code automatically generated by gen.go.
```
harryrackmil, line 2: we were going to get rid of these comments right? or at least update them to say "initially generated" or something

apps/lib/data_provider/gen.go
```diff
@@ -0,0 +1,273 @@
...
+	"github.com/rs/zerolog/pkgerrors"
+	"github.com/spf13/cobra"
+)
+
+const (
+	dirMode                  = 0o777
+	templatesDir             = "apps/lib/data_provider/configs/resources/templates"
+	sourceDir                = "apps/lib/data_provider/sources"
+	pathPrefix               = "// @path: "
+	importSourcesFile        = "apps/lib/data_provider/configs/import_sources.go"
```
harryrackmil, line 24: these are unused now right?

Makefile
```diff
@@ -14,3 +14,18 @@ help:
 		sub(/:.*/, "", $$1); \
 		printf "  %-30s %s\n", $$1, helpMessage; \
 	}' $(MAKEFILE_LIST)
+
+## Install the stork-data-provider binary
+.PHONY: install-data-provider-cli
+install-data-provider-cli:
+	@echo "Installing stork-data-provider..."
+	@echo "Running: go build -o $(shell go env GOPATH)/bin/stork-data-provider ./apps/cmd/data_provider"
+	@go build -o $(shell go env GOPATH)/bin/stork-data-provider ./apps/cmd/data_provider
```
akawalsky, line 23: Maybe it would make sense for us to build the package as part of ci and then host a binary in github instead?

ACK101101: harry what do you think? im not 100% sure what this means

harryrackmil: ah that would be nice if it's not too hard

harryrackmil: actually what's the flow when we make changes to code and want to run with the latest code? if users are iterating by running with the CLI it's going to be annoying

PR #79, opened 2025-01-22
Add data provider integration test

harryrackmil:
# Change
* Add a new `data_source_integration_test.go` file to both sources
    * integration tests just run the data source object in isolation and log n updates, failing if they don't receive any updates within some timeout interval. Will mostly be used for manually looking at data source output in the IDE
    * integration test files have `integration` build tag so they don't run in CI 
* add a template for these integration tests
* add a cancellable context to RunDataSource so the goroutines won't hang in tests
* 

# Testing
- [X] Ran integration tests in Goland and got useful output (and no failure)
- [X] CI test action doesn't run these integration tests
- [X] New generated sources compile correctly


apps/lib/data_provider/configs/source_config_tests/uniswapv2_test.go
```diff
@@ -20,11 +20,10 @@ func TestValidUniswapV2Config(t *testing.T) {
 			  	"dataSource": "uniswapv2",
 				"updateFrequency": "5s",
 				"contractAddress": "0x0d4a11d5eeaac28ec3f61d100daf4d40471f1852",
-				"httpProviderUrl": "https://eth-mainnet.g.alchemy.com/v2/",
-				"providerApiKeyEnvVar": "ALCHEMY_API_KEY",
-				"baseTokenIndex": 1,
+				"httpProviderUrl": "https://ethereum-rpc.publicnode.com",
+				"baseTokenIndex": 0,
```
harryrackmil, line 24: `1` and `2` for the indices here is a little weird since they're zero-indexed. It's a valid config since a pool could have more than 2 coins technically, but should probably use a more likely config

apps/lib/data_provider/configs/source_config_tests/uniswapv2_test.go
```diff
@@ -20,11 +20,10 @@ func TestValidUniswapV2Config(t *testing.T) {
 			  	"dataSource": "uniswapv2",
 				"updateFrequency": "5s",
 				"contractAddress": "0x0d4a11d5eeaac28ec3f61d100daf4d40471f1852",
-				"httpProviderUrl": "https://eth-mainnet.g.alchemy.com/v2/",
-				"providerApiKeyEnvVar": "ALCHEMY_API_KEY",
-				"baseTokenIndex": 1,
+				"httpProviderUrl": "https://ethereum-rpc.publicnode.com",
```
harryrackmil, line 23: use a valid url which doesn't require an API key

apps/lib/data_provider/sources/random/data_source_integration_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"testing"
+	"time"
+
+	"github.com/Stork-Oracle/stork-external/apps/lib/data_provider/types"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestRandomDataSource_RunDataSource(t *testing.T) {
```
harryrackmil, line 14: moved this from the random data source unit test - for pretty much every source except for random running the data source will hit external services, so makes sense to just put this in the integration test file

apps/lib/data_provider/sources/random/data_source_integration_test.go
```diff
@@ -0,0 +1,39 @@
...
+			MaxValue:        105.0,
+		},
+	}
+	dataSource := newRandomDataSource(config)
+	updateCh := make(chan types.DataSourceUpdateMap)
+	go dataSource.RunDataSource(updateCh)
+
+	// print a few messages, fail if none come through within the timeout period
+	numMessages := 10
+	timeoutDuration := 100 * time.Millisecond
```
harryrackmil, line 30: these numbers may need to be tweaked for different sources (for instance 100ms timeout between messages is probably not what we want for Uniswap)

apps/lib/data_provider/sources/random/data_source_integration_test.go
```diff
@@ -0,0 +1,39 @@
...
+	for i := 0; i < numMessages; i++ {
+		select {
+		case result := <-updateCh:
+			t.Logf("received update: %v", result)
```
harryrackmil, line 34: in the unit test version of this we were also checking that the value was within a valid range. Again, that doesn't make as much sense in the general case since it's hard to test whether a price is "correct" without needing to change your expected range frequently. Users will just inspect the output manually to check if prices look reasonable.

apps/lib/data_provider/sources/uniswapv2/data_source.go
```diff
@@ -15,7 +14,7 @@ import (
 	"github.com/rs/zerolog"
 )
 
-const uniswapV2AbiFileName = "random.json"
+const uniswapV2AbiFileName = "uniswap_v2.json"
```
harryrackmil, line 17: this was a find/replace mistake 🙃

apps/lib/data_provider/sources/uniswapv2/data_source_integration_test.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestUniswapDataSource_RunDataSource(t *testing.T) {
+	config := types.DataProviderSourceConfig{
+		Id: "WETHUSDT",
+		Config: UniswapV2Config{
+			DataSource:         UniswapV2DataSourceId,
+			UpdateFrequency:    "5s",
+			HttpProviderUrl:    "https://ethereum-rpc.publicnode.com",
```
harryrackmil, line 20: public rpc provider should be fine for integration testing at this rate

akawalsky (commented):
Couple comments here

apps/lib/data_provider/sources/uniswapv2/data_source_integration_test.go
harryrackmil: the intention is that they are skipped - they won't run as part of CI, only if a user runs them from their IDE or with the `integration` build tag

apps/lib/data_provider/configs/resources/templates/source/data_source_integration_test.gotmpl
```diff
@@ -0,0 +1,46 @@
+// @path: /apps/lib/data_provider/sources/{{ .LowerStr }}/data_source_integration_test.go
```
harryrackmil, line 1: add a template with a basic integration test that just runs the users data source against a real external data source until it gets 10 updates. Generated this code from the template using the `stork-generate` cli:
```
//go:build integration
// +build integration
// This file contains integration tests for this data source.


package harry

import (
	"context"
	"testing"
	"time"

	"github.com/Stork-Oracle/stork-external/apps/lib/data_provider/types"
	"github.com/stretchr/testify/assert"
)

// This test will hit real external data sources. It's meant to be run manually so you can manually examine the results.
func TestHarryDataSource_RunDataSource(t *testing.T) {
	config := types.DataProviderSourceConfig{
		Id: "MY_TEST_VALUE_ID",
		Config: HarryConfig{
			DataSource:      HarryDataSourceId,
			// TODO: add valid configuration
		},
	}

	dataSource := newHarryDataSource(config)
	updateCh := make(chan types.DataSourceUpdateMap)
	ctx, cancel := context.WithCancel(context.Background())
	defer cancel()
	go dataSource.RunDataSource(ctx, updateCh)

	// print a few messages, fail if none come through within the timeout period
	numMessages := 10
	timeoutDuration := 10 * time.Second // TODO: change this timeout if needed

	for i := 0; i < numMessages; i++ {
		select {
		case result := <-updateCh:
			t.Logf("received update: %v", result)
		case <-time.After(timeoutDuration):
			assert.Fail(t, "didn't receive update from data source in time")
		}
	}
}
```

apps/lib/data_provider/configs/resources/templates/source/data_source_test.gotmpl
```diff
@@ -1,6 +1,6 @@
 // @path: /apps/lib/data_provider/sources/{{ .LowerStr }}/data_source_test.go
 // Code initially generated by gen.go.
-// This file contains unit tests for functions implemented in the data source puller.
+// This file contains unit tests for functions implemented in the data source.
```
harryrackmil, line 3: small change - was trying to avoid `puller` since the data sources aren't necessarily pulling (they could be receiving push data)

apps/lib/data_provider/runner.go
```diff
@@ -34,8 +36,12 @@ func (r *DataProviderRunner) Run() {
 	if err != nil {
 		panic("unable to build data sources: " + err.Error())
 	}
+
+	ctx, cancel := context.WithCancel(context.Background())
```
harryrackmil, line 40: make data source go routines cancellable

apps/lib/data_provider/sources/random/data_source.go
```diff
@@ -36,14 +37,14 @@ func newRandomDataSource(sourceConfig types.DataProviderSourceConfig) *randomDat
 	}
 }
 
-func (r randomDataSource) RunDataSource(updatesCh chan types.DataSourceUpdateMap) {
+func (r randomDataSource) RunDataSource(ctx context.Context, updatesCh chan types.DataSourceUpdateMap) {
```
harryrackmil, line 40: `ctx` is now part of the data source interface so we can make sure data source goroutines don't hang indefinitely

apps/lib/data_provider/sources/scheduler.go
```diff
@@ -25,10 +26,18 @@ func NewScheduler(
...
-		s.emitUpdate(updatesCh)
+	ticker := time.NewTicker(s.updateFrequency)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ticker.C:
+			s.emitUpdate(updatesCh)
+		case <-ctx.Done():
+			return
```
harryrackmil, line 39: here's where we actually respect the context cancellation

apps/lib/data_provider/sources/random/data_source_integration_test.go
```diff
@@ -0,0 +1,39 @@
...
+		Config: RandomConfig{
+			DataSource:      RandomDataSourceId,
+			UpdateFrequency: "50ms",
+			MinValue:        101.0,
+			MaxValue:        105.0,
+		},
+	}
+	dataSource := newRandomDataSource(config)
+	updateCh := make(chan types.DataSourceUpdateMap)
+	go dataSource.RunDataSource(updateCh)
```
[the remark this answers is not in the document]
harryrackmil: yea good idea - added a context object with cancel to handle this

apps/lib/data_provider/configs/resources/templates/source/data_source_integration_test.gotmpl
ACK101101: nice!

akawalsky (approved):
Looking really clean, love it

apps/lib/data_provider/sources/scheduler.go
ACK101101: just confirming that we want the runner to pass the same context to each data source. 
this makes sense because we only intend on developers adding / running one data source at a time?

ACK101101 (approved):

apps/lib/data_provider/sources/scheduler.go
harryrackmil: talked about this offline - the only situation where all data sources would get cancelled is if we call the `cancel()` function explicitly (which only happens from the defer here, which is the desired behavior). Individual data sources won't be allowed to call cancel since they only have the context, not the cancel function.

PR #82, opened 2025-01-28
Add Raydium CLMM as Data Source Using CLI Tool

ACK101101:


apps/lib/data_provider/configs/data_source_config.json
```diff
@@ -0,0 +1,37 @@
+{
```
harryrackmil, line 1: let's not commit this config - in general we'll expect users to pass their own config file which doesn't necessarily live in the repo

harryrackmil (approved):
Looks good!

PR #84, opened 2025-01-29
QOL Improvements to Data Provider

ACK101101:
Separated the cli into two parts, stork-generate and data-provider. This is so the user doesn't have to reinstall/recompile anything related to code generation. When iterating, they only have to run the make start-data-provider command. 
Added remove source command for convenience. Removed the need for flags when running generate or remove.
Use public node for solana so no api key is needed for raydiumclmm.
Updated readme.
Add the animation back! Runs only when installing stork-generate, which should only happen once.


apps/lib/generate/gen.go
```diff
@@ -280,55 +280,55 @@ func pascalToCamel(pascalName string) string {
 	return strings.ToLower(pascalName[:endOfFirstWord]) + pascalName[endOfFirstWord:]
 }
 
-func removeDataProvider(cmd *cobra.Command, args []string) error {
-	dataProviderName, _ := cmd.Flags().GetString(DataProviderNameFlag)
+// func removeDataProvider(cmd *cobra.Command, args []string) error {
```
akawalsky, line 283: Whats going on here? We should probably either include this or remove it rather than comment it out

ACK101101: yeah this was just temporary to get around a merge conflict, i think i forgot to push up the version where it is included

akawalsky (approved):
:shipit: once you take care of commented out code

akawalsky (approved):

apps/lib/data_provider/configs/resources/templates/source/config.gotmpl
```diff
@@ -1,5 +1,6 @@
 // @path: /apps/lib/data_provider/sources/{{ .LowerStr }}/config.go
 // Code initially generated by gen.go.
+// This file defines the configuration for the data source.
```
harryrackmil, line 3: nice

apps/lib/generate/command.go
```diff
@@ -0,0 +1,41 @@
...
+	Short: "Generate a new data provider",
+	Args:  cobra.ExactArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		return generateDataProvider(cmd, args)
+	},
+}
+
+var RemoveCmd = &cobra.Command{
+	Use:   "remove [name]",
+	Short: "Remove code related toa data source integration",
```
harryrackmil, line 21: `to a`

PR #86, opened 2025-01-29
Henry/sto 690 a series of small fixes

HenryMBaldwin:
This contains a series of small fixes within stork-external.

The biggest of these fixes renaming stork-sdk to stork-solana-sdk. This involved changes to all of our solana packages in this repo, multiple readmes, the official docs, as well as publishing a new crate to cratesio and deprecating the old one.

The solana change will require a redeploy on both mainnet and devnet.


akawalsky (approved):

HenryMBaldwin:
Turns out if we redeploy with the new crate it shouldn't break dependant contracts unless there's a breaking change in the way the structs are serialized, which upon testing it appears a crate name change does not break this. Due to this, we don't even really need to redeploy at all. I still want to redeploy eventually, however I'll look into doing a verified deploy so it can be a future task.

PR #93, opened 2025-02-04
Add transformations

akawalsky:


apps/lib/data_provider/transformations/parse.go
```diff
@@ -0,0 +1,375 @@
...
+	"github.com/alecthomas/participle/v2"
+	"gonum.org/v1/gonum/graph"
+	"gonum.org/v1/gonum/graph/simple"
+	"gonum.org/v1/gonum/graph/topo"
+)
+
+// heavily borrowed from https://github.com/alecthomas/participle/blob/master/_examples/expr/main.go
+
+type evaluable interface {
+	Eval(ctx map[string]types.DataSourceValueUpdate) float64
```
harryrackmil, line 19: naming this something like `updateMap` might be more explicit so it's clear we're not passing a `context.Context` object

apps/lib/data_provider/configs/resources/data_provider_config.schema.json
```diff
@@ -29,6 +29,22 @@
           "config"
         ]
       }
+    },
+    "transformations": {
+      "type": "array",
+      "description": "A list of transformations to apply to the data",
+      "minItems": 1,
```
harryrackmil, line 36: do we care about preventing users from passing an empty list of tranformations? seems fine naively

apps/lib/data_provider/transformations/parse_test.go
```diff
@@ -0,0 +1,176 @@
...
+		t.Run(test.name, func(t *testing.T) {
+			expr, err := parse(test.formula)
+			require.NoError(t, err)
+			out := expr.Eval(test.variables)
+			require.Equal(t, test.expected, out)
+		})
+	}
+}
+
+func TestGetDependencies(t *testing.T) {
```
harryrackmil, line 95: let's test our cycle detection logic too

harryrackmil (commented):
this is looking awesome!

apps/lib/data_provider/transformations/parse_test.go
```diff
@@ -0,0 +1,93 @@
...
+			expected: 6.0,
+		},
+		{
+			name:     "nested expression",
+			formula:  "median(1, 2, 3) * mean(4, 5, 6)",
+			expected: 10.0,
+		},
+		{
+			name:    "nested expression with variables",
+			formula: "median(x, y, z) * mean(4, 5, 6)",
```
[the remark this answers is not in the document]
akawalsky: Yeah definitely. Product and sum are just helper funcs, we could definitely remove them if they're confusing

apps/lib/data_provider/configs/resources/data_provider_config.schema.json
akawalsky: You could make a similar argument for sources now that transformations is here, but happy to remove this restriction

apps/lib/data_provider/transformations/parse_test.go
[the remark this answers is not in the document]
harryrackmil: no I like having both I think - with longer lists `sum(a, b, c, d, e)` is a little nicer than `a + b + c + d + e` I think

apps/lib/data_provider/configs/resources/data_provider_config.schema.json
harryrackmil: ah I guess that's true - a user could just a transformation of `1 + 2` or something and not have any sources. I kinda like the idea of removing this minItems restriction from both? e.g. if a user were auto-generating this config it's not crazy that one of them might intentionally be an empty list.

harryrackmil (dismissed):

harryrackmil (approved):

PR #96, opened 2025-02-04
[Data Provider] Only update transformations on source value change

harryrackmil:
# Change
A transformation is a pure function of its input values, so there's no need to report a transformation's value unless one of its inputs has changed. We can accomplish this by regularly recalculating all the transformations and checking if the output has changed, but that requires a lot of unnecessary calculation and adds some latency between checks.

Instead, we can wait for a source value update and immediately recalculate only the affected transformations, to minimize compute and latency.

Also got rid of the namespacing concept (`s.var1` and `t.var1`) and started enforcing uniqueness of IDs.

# Testing
Ran locally with the updated example config and confirmed that updates were as expected, and only changed transformations came through in each update:
```
{"level":"info","application":"stork-data-provider","service":"main","time":"2025-02-04T15:06:25.404693-08:00","message":"Starting data provider"}
{"level":"debug","application":"stork-data-provider","service":"writer","time":"2025-02-04T15:06:25.406541-08:00","message":"Update: {\"type\":\"prices\",\"data\":[{\"t\":1738710385406382000,\"a\":\"MY_RANDOM_VALUE\",\"v\":\"2881.177269307398546516\"},{\"t\":1738710385406435000,\"a\":\"MY_RANDOM_VALUE_SQUARED\",\"v\":\"8301182.457173638045787811\"}]}"}
{"level":"debug","application":"stork-data-provider","service":"writer","time":"2025-02-04T15:06:25.729675-08:00","message":"Update: {\"type\":\"prices\",\"data\":[{\"t\":1738710385729000000,\"a\":\"SOL_USDC\",\"v\":\"205.494210851040492116\"},{\"t\":1738710385729654000,\"a\":\"USDC_SOL\",\"v\":\"0.004866317137882216\"}]}"}
{"level":"debug","application":"stork-data-provider","service":"writer","time":"2025-02-04T15:06:25.811789-08:00","message":"Update: {\"type\":\"prices\",\"data\":[{\"t\":1738710385811000000,\"a\":\"WETHUSDT\",\"v\":\"2684.571143644646781468\"},{\"t\":1738710385811750000,\"a\":\"SOL_WETH_MEDIAN\",\"v\":\"1445.032677247843594159\"}]}"}
{"level":"debug","application":"stork-data-provider","service":"writer","time":"2025-02-04T15:06:26.407528-08:00","message":"Update: {\"type\":\"prices\",\"data\":[{\"t\":1738710386407480000,\"a\":\"MY_RANDOM_VALUE_SQUARED\",\"v\":\"6326522.123877909034490585\"},{\"t\":1738710386407471000,\"a\":\"MY_RANDOM_VALUE\",\"v\":\"2515.257864291037094517\"}]}"}
{"level":"debug","application":"stork-data-provider","service":"writer","time":"2025-02-04T15:06:27.40759-08:00","message":"Update: {\"type\":\"prices\",\"data\":[{\"t\":1738710387407495000,\"a\":\"MY_RANDOM_VALUE\",\"v\":\"2798.393465372781975020\"},{\"t\":1738710387407501000,\"a\":\"MY_RANDOM_VALUE_SQUARED\",\"v\":\"7831005.987041087821125984\"}]}"}
{"level":"debug","application":"stork-data-provider","service":"writer","time":"2025-02-04T15:06:28.407614-08:00","message":"Update: {\"type\":\"prices\",\"data\":[{\"t\":1738710388407527000,\"a\":\"MY_RANDOM_VALUE\",\"v\":\"2974.764658530802535097\"},{\"t\":1738710388407536000,\"a\":\"MY_RANDOM_VALUE_SQUARED\",\"v\":\"8849224.773643882945179939\"}]}"}
{"level":"debug","application":"stork-data-provider","service":"writer","time":"2025-02-04T15:06:29.407714-08:00","message":"Update: {\"type\":\"prices\",\"data\":[{\"t\":1738710389407528000,\"a\":\"MY_RANDOM_VALUE\",\"v\":\"2527.944062149921592209\"},{\"t\":1738710389407537000,\"a\":\"MY_RANDOM_VALUE_SQUARED\",\"v\":\"6390501.181359047070145607\"}]}"}
{"level":"debug","application":"stork-data-provider","service":"writer","time":"2025-02-04T15:06:30.4069-08:00","message":"Update: {\"type\":\"prices\",\"data\":[{\"t\":1738710390406743000,\"a\":\"MY_RANDOM_VALUE\",\"v\":\"2754.500418097115471028\"},{\"t\":1738710390406755000,\"a\":\"MY_RANDOM_VALUE_SQUARED\",\"v\":\"7587272.553297183476388454\"}]}"}
{"level":"debug","application":"stork-data-provider","service":"writer","time":"2025-02-04T15:06:30.832703-08:00","message":"Update: {\"type\":\"prices\",\"data\":[{\"t\":1738710390832644000,\"a\":\"USDC_SOL\",\"v\":\"0.004867169020998824\"},{\"t\":1738710390832644000,\"a\":\"SOL_WETH_MEDIAN\",\"v\":\"1445.014693790927594819\"},{\"t\":1738710390832000000,\"a\":\"SOL_USDC\",\"v\":\"205.458243937208351326\"}]}"}
{"level":"debug","application":"stork-data-provider","service":"writer","time":"2025-02-04T15:06:31.183649-08:00","message":"Update: {\"type\":\"prices\",\"data\":[{\"t\":1738710391183000000,\"a\":\"WETHUSDT\",\"v\":\"2685.236325866502284043\"},{\"t\":1738710391183575000,\"a\":\"SOL_WETH_MEDIAN\",\"v\":\"1445.347284901855346106\"}]}"}
```


apps/lib/data_provider/configs/example_data_source_config.json
```diff
@@ -31,5 +31,19 @@
         "maxValue": 3000
       }
     }
+  ],
+  "transformations": [
```
harryrackmil, line 35: toy examples which I used for my testing - not sure if we want to keep them in the example config or not

apps/lib/data_provider/runner.go
```diff
@@ -54,45 +52,17 @@ func (r *DataProviderRunner) Run() {
...
-
-				// Only add to updateMap if value has changed
-				if existing, ok := currentVals["t."+string(transformation.Id)]; !ok || existing.Value != computed.Value {
-					updateMap[transformation.Id] = computed
-				}
-			}
-			if len(updateMap) > 0 {
-				r.outputCh <- updateMap
-			}
+		case sourceUpdates := <-r.updatesCh:
```
harryrackmil, line 63: no scheduled tasks - we only act when there's a change to an input value.

Also note that we don't need to check if the update map is empty since there will at minimum be a source update

apps/lib/data_provider/transformations/parse.go
```diff
@@ -303,21 +304,92 @@ func parse(formula string) (*Expression, error) {
 	return parser.ParseString("", formula)
 }
 
-type OrderedTransformation struct {
-	Id             types.ValueId
-	Transformation *Expression
+const TransformationDataSourceId = types.DataSourceId("transformation")
```
harryrackmil, line 307: all updates for transformation values will have the DataSourceId `transformation` - kinda arbitrary, but nice to be able to group together transformations if we wanted to.

apps/lib/data_provider/transformations/parse.go
```diff
@@ -303,21 +304,92 @@ func parse(formula string) (*Expression, error) {
...
 }
 
-func (o *OrderedTransformation) String() string {
-	return fmt.Sprintf("%s: %s", o.Id, o.Transformation.String())
+func (tg *TransformationGraph) ProcessSourceUpdates(sourceUpdates types.DataSourceUpdateMap) types.DataSourceUpdateMap {
+	finalUpdateMap := make(types.DataSourceUpdateMap)
+
+	updateTime := time.Now()
+
+	// do a breadth-first traversal get all affected nodes
```
harryrackmil, line 340: two step algorithm to update affected prices:
1. Find a set of "dirty nodes" which need to be updated. We do this using a breadth first search which begins with all of the value ids included in the source update.
2. Iterate through all the nodes in the graph in topological order but only recalculate nodes which are in the "dirty nodes" set. This ensures we update all affected nodes, and that we update them in the correct order.

Note that we can't just do a breadth first search and update as we traverse because we could have a graph like this:
source1 ---> var 1 -> var2 -> var3
source2 -------------------> var3

breadth first search would update var3 before updating var2, so we'd wind up with inaccurate values

apps/lib/data_provider/transformations/parse.go
```diff
@@ -303,21 +304,92 @@ func parse(formula string) (*Expression, error) {
...
+		currentVals:           make(map[string]types.DataSourceValueUpdate),
+	}
 }
 
-func (o *OrderedTransformation) String() string {
-	return fmt.Sprintf("%s: %s", o.Id, o.Transformation.String())
+func (tg *TransformationGraph) ProcessSourceUpdates(sourceUpdates types.DataSourceUpdateMap) types.DataSourceUpdateMap {
+	finalUpdateMap := make(types.DataSourceUpdateMap)
+
+	updateTime := time.Now()
```
harryrackmil, line 338: It would be kinda nice if the timestamp on each transformation was some function of the timestamp of its inputs (e.g. the max of all source prices it depends on or something), but that felt a little complicated and not incredibly valuable. For now I'm just setting all transformations in a batch to the time that the batch started, so it should be pretty close to the update timestamp.

apps/lib/data_provider/transformations/parse.go
```diff
@@ -303,21 +304,92 @@ func parse(formula string) (*Expression, error) {
...
+
+	// do a breadth-first traversal get all affected nodes
+	dirtyTransformationNodes := make(map[graph.Node]interface{})
+	queue := make([]graph.Node, 0)
+	for valueId, sourceUpdate := range sourceUpdates {
+		queue = append(queue, tg.valueIdToNode[valueId])
+		finalUpdateMap[valueId] = sourceUpdate
+		tg.currentVals[string(valueId)] = sourceUpdate
```
harryrackmil, line 346: We can have multiple source updates in a single `DataSourceUpdateMap`, so we need to add all of them to the BFS queue (so really it's a multi-source breadth first traversal).

Also note that we're maintaining two very similar maps here. The final update map is scoped to this function call and just holds changed values that we want to report. The current value map is long-lived state - we need it to compute transformations involving multiple inputs in situations where only one input may have changed.

apps/lib/data_provider/transformations/parse.go
```diff
@@ -303,21 +304,92 @@ func parse(formula string) (*Expression, error) {
...
+	}
+
+	for len(queue) > 0 {
+		current := queue[0]
+		queue = queue[1:]
+
+		nodeIterator := tg.dependencyGraph.From(current.ID())
+		for nodeIterator.Next() {
+			nextNode := nodeIterator.Node()
+			if _, seen := dirtyTransformationNodes[nextNode]; !seen {
```
harryrackmil, line 356: don't traverse nodes twice

apps/lib/data_provider/transformations/parse.go
```diff
@@ -303,21 +304,92 @@ func parse(formula string) (*Expression, error) {
...
+		}
+	}
+
+	// update dirty transformations in topological order
+	for _, node := range tg.orderedNodes {
+		if _, isDirty := dirtyTransformationNodes[node]; isDirty {
+			transformationValueId := tg.nodeToValueId[node]
+			transformation := tg.parsedTransformations[transformationValueId]
+			transformationValue := transformation.Eval(tg.currentVals)
+			if math.IsNaN(transformationValue) {
```
harryrackmil, line 369: if any of the inputs are missing, don't report the output. We'll just wait for all inputs to be present.

apps/lib/data_provider/transformations/parse.go
```diff
@@ -329,8 +401,21 @@ func BuildTransformations(transformations []types.DataProviderTransformationConf
...
-		nodeToTransformationId[node] = transformation.Id
-		transformationIdToNode[transformation.Id] = node
+		nodeToValueId[node] = transformation.Id
+		if _, exists := valueIdToNode[transformation.Id]; exists {
+			return nil, fmt.Errorf("duplicate value id: %v", transformation.Id)
+		}
+		valueIdToNode[transformation.Id] = node
+	}
+
+	for sourceId := range sourceIds {
```
harryrackmil, line 411: make source ids nodes too

apps/lib/data_provider/transformations/parse.go
```diff
@@ -329,8 +401,21 @@ func BuildTransformations(transformations []types.DataProviderTransformationConf
...
+		node := g.NewNode()
+		g.AddNode(node)
+		nodeToValueId[node] = sourceId
+		if _, exists := valueIdToNode[sourceId]; exists {
```
harryrackmil, line 415: enforce uniqueness of ids - we weren't doing this between source ids previously. Now we're enforcing it globally, so a user can't define a transformation id that conflicts with a source id

apps/lib/data_provider/transformations/parse.go
```diff
@@ -341,35 +426,24 @@ func BuildTransformations(transformations []types.DataProviderTransformationConf
...
-				if _, ok := transformationIdToNode[types.ValueId(dep)]; !ok {
-					return nil, fmt.Errorf("no such transformation: %s", dep)
-				}
-				g.SetEdge(g.NewEdge(transformationIdToNode[types.ValueId(dep)], transformationIdToNode[transformation.Id]))
-			} else if strings.HasPrefix(dep, "s.") {
-				dep = dep[2:]
-				if _, ok := sourceIds[types.ValueId(dep)]; !ok {
-					return nil, fmt.Errorf("no such source: %s", dep)
+			_, sourcePriceExists := sourceIds[types.ValueId(dep)]
+			if !sourcePriceExists {
```
harryrackmil, line 430: we don't really need to handle sources and transformations differently - we just need to fail if we have neither a source nor a transformation for the relevant dependency

apps/lib/data_provider/transformations/parse.go
```diff
@@ -341,35 +426,24 @@ func BuildTransformations(transformations []types.DataProviderTransformationConf
...
+				_, transformationExists := valueIdToNode[types.ValueId(dep)]
+				if !transformationExists {
+					return nil, fmt.Errorf("no such source or transformation id: %s", dep)
 				}
-			} else {
-				return nil, fmt.Errorf("unknown dependency: %s", dep)
 			}
+
+			g.SetEdge(g.NewEdge(valueIdToNode[types.ValueId(dep)], valueIdToNode[transformation.Id]))
```
harryrackmil, line 437: add edges from dependency to transformation (regardless of whether the dependency is a source or transformation)

apps/lib/data_provider/transformations/parse_test.go
```diff
@@ -136,41 +138,222 @@ func TestBuildTransformations(t *testing.T) {
 			transformations: []types.DataProviderTransformationConfig{
 				{
 					Id:      "vtest1",
-					Formula: "s.test1 + 2",
```
harryrackmil, line 139: get rid of namespacing

apps/lib/data_provider/transformations/parse_test.go
```diff
@@ -136,41 +138,222 @@ func TestBuildTransformations(t *testing.T) {
...
+					Formula: "test1 + 2",
 				},
 			},
 			knownSources: map[types.ValueId]interface{}{
 				"test1": nil,
 			},
-			expectedOrder: []types.ValueId{"vtest1"},
+			expectedOrder: []types.ValueId{"test1", "vtest1"},
```
harryrackmil, line 147: now that source nodes are in the graph too, the expected order includes the source node

apps/lib/data_provider/transformations/parse_test.go
```diff
@@ -136,41 +138,222 @@ func TestBuildTransformations(t *testing.T) {
...
+		{
+			name: "duplicate value ids",
+			transformations: []types.DataProviderTransformationConfig{
+				{
+					Id:      "test1",
+					Formula: "test1 + 2",
+				},
+			},
+			knownSources: map[types.ValueId]interface{}{
+				"test1": nil,
```
harryrackmil, line 197: don't allow a transformation with the same id as a source

apps/lib/data_provider/transformations/parse_test.go
```diff
@@ -136,41 +138,222 @@ func TestBuildTransformations(t *testing.T) {
...
+	outgoingUpdates := transformationGraph.ProcessSourceUpdates(map[types.ValueId]types.DataSourceValueUpdate{
+		"test1": {
+			ValueId:      "test1",
+			DataSourceId: "",
+			Timestamp:    time.Now(),
+			Value:        10.0,
+		},
+	})
+
+	assert.Len(t, outgoingUpdates, 2)
```
harryrackmil, line 276: even though `test1` affects `multiple_input`, we don't output `multiple_input` because we haven't yet received a value for `test2`

apps/lib/data_provider/transformations/parse.go
```diff
@@ -303,21 +304,92 @@ func parse(formula string) (*Expression, error) {
 	return parser.ParseString("", formula)
 }
 
-type OrderedTransformation struct {
-	Id             types.ValueId
-	Transformation *Expression
+const TransformationDataSourceId = types.DataSourceId("transformation")
+
+type TransformationGraph struct {
```
akawalsky, line 309: It probably makes sense to move this to a separate file and so parse and graph related stuff can be a little dissociated

akawalsky (approved):
This looks good, just a comment about moving some code around.

Also, it might be nice to see a benchmark of a long path of transformations being run via the graph traversal vs the corollary raw transformations being run directly in code to get a sense of how much latency is added in the graph traversal.

apps/lib/data_provider/transformations/graph_test.go
```diff
@@ -0,0 +1,291 @@
...
+		},
+	})
+
+	assert.Len(t, outgoingUpdates, 1)
+	update, exists = outgoingUpdates["test3"]
+	assert.True(t, exists)
+	assert.Equal(t, 1.0, update.Value)
+}
+
+func BenchmarkTranformationGraph_ProcessSourceUpdates(b *testing.B) {
```
harryrackmil, line 246: Test graph traversal + calculation time by creating two long linear dependency chains (100 transformation nodes) where the first chain depends on source price `test_source_a` and the second depends on `test_source_b`.

When `test_source_a` changes, doing the graph traversal and calculating the new values for all 100 relevant nodes takes about 40 microseconds (0.04 milliseconds), and when I edited the benchmark locally to update both source prices on every iteration (meaning we needed to traverse + update 200 nodes) it took about 80 microseconds.

When I increased the length of both chains to 1000 transformations and updated `test_source_a`, it took about 500 micros, so only half a millisecond.

akawalsky (approved):

PR #101, opened 2025-02-13
fix: Get libwasmvm in dockerfile.

HenryMBaldwin:


Dockerfile
```diff
@@ -43,9 +43,15 @@ RUN apt-get update && \
         ca-certificates \
         curl \
         jq \
+        wget \
```
akawalsky, line 46: Do we need wget or can we get away with using curl?

Dockerfile
```diff
@@ -43,9 +43,15 @@ RUN apt-get update && \
...
     && rm -rf /var/lib/apt/lists/* \
     && apt-get clean
 
+# Install CosmWasm libraries
+RUN wget https://github.com/CosmWasm/wasmvm/releases/download/v2.2.1/libwasmvm.aarch64.so -O /usr/local/lib/libwasmvm.aarch64.so && \
+    wget https://github.com/CosmWasm/wasmvm/releases/download/v2.2.1/libwasmvm.x86_64.so -O /usr/local/lib/libwasmvm.x86_64.so && \
+    cp "/usr/local/lib/libwasmvm.$(uname -m | sed 's/x86_64/x86_64/;s/aarch64/aarch64/').so" /usr/local/lib/libwasmvm.so
```
akawalsky, line 53: Can this just go in usr local lib to avoid the extra option below?

akawalsky (commented):
Does run local in makefile also need to be adapted?

Dockerfile
HenryMBaldwin: already changed this lol

akawalsky (approved):

HenryMBaldwin:
> Does run local in makefile also need to be adapted?

Yes, I moved the install to the build make install command. Now works for both local builds and in docker.

PR #107, opened 2025-03-18
Cronos zkevm l2 mainnet deployment.

HenryMBaldwin:


contracts/evm/hardhat.config.ts
```diff
@@ -350,7 +359,7 @@ const config: HardhatUserConfig = {
     }
   },
   etherscan: {
-    // enabled: false, // uncomment this for ZKSync verifications
+    // enabled: true, // uncomment this for ZKSync verifications
```
harryrackmil, line 362: this is intentional? if we're doing a ZKSync verification we want this to be set to true in the future?

harryrackmil (dismissed):

contracts/evm/hardhat.config.ts
HenryMBaldwin: Oh yeah that change was left over from me trying to get verification working (I gave up). Changed back

harryrackmil (approved):

akawalsky (commented):
I think this also requires a change to evm/docker-entrypoint and lambdas-node.dockerfile

PR #109, opened 2025-03-24
Add cronos l2 api key to docker-entrypoint.sh

HenryMBaldwin:
Needed to add var from cronos deployment in 1 more spot


HenryMBaldwin:
> Also add to lambdas-node.Dockerfile

https://github.com/Stork-Oracle/stork-aggregator/pull/682

akawalsky (approved):

PR #112, opened 2025-04-08
chore: Simplified Hardhat Node Deployment Script

0xwitty:
I’ve removed the port check and Hardhat node waiting logic to simplify the script. This makes it cleaner, assuming the node will be ready by the time deployment starts. I kept `wait $!` to ensure the script doesn’t finish before the node is done.

### DIFF
--- a/contracts/evm/docker-entrypoint.sh
+++ b/contracts/evm/docker-entrypoint.sh
@@ -1,28 +1,29 @@
 #!/bin/sh
 
 # Change to the correct directory
-cd /usr/src/app;
+cd /usr/src/app
 
-# this is a fake private key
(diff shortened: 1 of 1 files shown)


contracts/evm/docker-entrypoint.sh
```diff
@@ -1,28 +1,29 @@
...
 
 # Start hardhat node as a background process
 npx hardhat node &
 
+# Compile contracts
 npx hardhat compile
 
 # Wait for hardhat node to initialize and then deploy contracts
-npx wait-on http://127.0.0.1:8545 && npx hardhat --network inMemoryNode deploy $STORK_PUBLIC_KEY
+npx hardhat --network inMemoryNode deploy $STORK_PUBLIC_KEY
```
akawalsky, line 26: You mention in description that there is still some wait logic. Where is that?

akawalsky (commented):
A few comments and questions here

akawalsky:
Closing due to inactivity

PR #116, opened 2025-04-24
Alexander/sto 801 pusher orchestrator implementation

ACK101101:


apps/lib/chain_pusher/solana.go
```diff
@@ -61,12 +68,6 @@ func NewSolanaContractInteracter(rpcUrl, wsUrl, contractAddr string, privateKeyF
 		return nil, err
 	}
 
-	payer, err := solana.PrivateKeyFromSolanaKeygenFile(privateKeyFile)
```
ACK101101, line 64: the function resolved string to []byte

ACK101101 (commented):
in summary:
renamed instances of Interacter to Interactor
renamed instances of InternalStorkStructsTemporalNumericValue to InternalTemporalNumericValue
added context to ListenContractEvents() to make it cancellable

code should not be merged as is, the change of privateKey / mnemonicFile from string to []byte is not tested. it was made to fit the expected format of the privateKey in the pusherOrchestrator

apps/lib/chain_pusher/utilities.go
```diff
@@ -71,13 +70,9 @@ func stringToByte32(input string) ([32]byte, error) {
 }
 
 // For simplicity, this function assumes the mnemonic file contains the private key directly
-func loadPrivateKey(mnemonicFile string) (*ecdsa.PrivateKey, error) {
-	data, err := os.ReadFile(mnemonicFile)
```
akawalsky, line 75: Can you pull this back up to the evm-push command?

akawalsky (commented):
Looking pretty good! Left a few cleanup comments. @harryrackmil can you review as well?

PR #121, opened 2025-05-15
error logging changes on evm pull

ACK101101:


apps/lib/chain_pusher/evm.go
```diff
@@ -137,9 +137,9 @@ func (sci *EvmContractInteractor) PullValues(encodedAssetIds []InternalEncodedAs
 		storkStructsTemporalNumericValue, err := sci.contract.GetTemporalNumericValueV1(nil, encodedAssetId)
 		if err != nil {
 			if strings.Contains(err.Error(), "NotFound()") {
-				sci.logger.Debug().Str("assetId", hex.EncodeToString(encodedAssetId[:])).Msg("No value found")
+				sci.logger.Warn().Err(err).Str("assetId", hex.EncodeToString(encodedAssetId[:])).Msg("No value found")
```
akawalsky, line 140: This seems like a bit of an antipattern. Seems like the PullValues allows for returning error, but is instead being opinionated about logging or not in the interactor rather than in the user of the interactor. Maybe instead we return some specific error types and allow for the consumer to decide?

apps/lib/chain_pusher/evm.go
```diff
@@ -137,9 +137,9 @@ func (sci *EvmContractInteractor) PullValues(encodedAssetIds []InternalEncodedAs
...
 			} else {
-				sci.logger.Debug().Str("assetId", hex.EncodeToString(encodedAssetId[:])).Msg("Failed to get latest value")
+				sci.logger.Warn().Err(err).Str("assetId", hex.EncodeToString(encodedAssetId[:])).Msg("Failed to get latest value")
```
akawalsky, line 142: It does seem like this should be error log, maybe we do the above and then threshold log in the consumer?

apps/lib/chain_pusher/evm.go
ACK101101: Yeah, that could make more sense. We don't actually return any real errors here anyway. Also, since it is operating in a for loop, we would only be able to capture one of the errors if multiple occur. Should we return a list of errors? Also, some of the existing pusher code throws Fatal if PullValues() returns an error.

akawalsky: Really? Doesnt seem like that should be a fatal error.

I see, that does make more sense actually. Eh maybe it is fine to just log here for now then.

akawalsky: Maybe the higher order thing can compare the size of the polled vals to the input map and see if there is a discrepancy and only log error using threshold logger then?

ACK101101: ok i like that

ACK101101: Actually, what about the case where we stop pushing an asset? That will still show up on successful pulls and `len(values)` could be > `len(encodedAssetIDs)` 
I'm leaning towards removing logging an error on the higher order PullValues call (since there are no real errors to be returned) and just threshold logging missing values (`Msg("no on-chain value found for encodedAssetID")`)

akawalsky: ok sounds good

PR #143, opened 2025-06-27
Asset Whitelist Bugfix

harryrackmil:
# Problem
When we removed an asset from the whiteslist for a publisher and the publisher agent received an update for an asset which was not whitelisted, we observed:
1. The publisher agent would sometimes send empty messages to the aggregator, which is treated as an invalid message and caused the aggregator to disconnect the publisher repeatedly
2. When disconnected, the existing `RunOutgoingConnection` goroutine would restart with the originally configured asset list, even though the asset list had changed in the registry

# Testing
- [x] adding a new asset to the whitelist is respected for a running agent
- [x] removing an asset from the whitelist is respected for a running agent
- [x] the agent reconnects successfully after an incoming websocket disconnect
- [x] the agent reconnect successfully after an outgoing websocket disconnect

### DIFF
--- a/apps/lib/publisher_agent/runner.go
+++ b/apps/lib/publisher_agent/runner.go
@@ -142,6 +142,7 @@ func (r *PublisherAgentRunner[T]) Run() {
 }
 
 func (r *PublisherAgentRunner[T]) RunOutgoingConnection(url BrokerPublishUrl, assetIds map[AssetId]struct{}) {
+	outgoingWebsocketConn := NewOutgoingWebsocketConnection[T](assetIds, r.logger)
 	for {
 		r.logger.Debug().Msgf("Connecting to receiver WebSocket with url %s", url)
 
@@ -169,7 +170,7 @@ func (r *PublisherAgentRunner[T]) RunOutgoingConnection(url BrokerPublishUrl, as
 				r.outgoingConnectionsLock.Unlock()
 			},
 		)
-		outgoingWebsocketConn := NewOutgoingWebsocketConnection[T](websocketConn, assetIds, r.logger)
+		outgoingWebsocketConn.SetWebsocketConn(websocketConn)
 
 		// add subscriber to list
 		r.outgoingConnectionsLock.Lock()
--- a/apps/lib/publisher_agent/runner.go
+++ b/apps/lib/publisher_agent/runner.go
@@ -76,6 +76,7 @@ func (r *PublisherAgentRunner[T]) UpdateBrokerConnections() {
 	}
 
 	// add or update desired connections
+	r.outgoingConnectionsLock.RLock()
 	for brokerUrl, newAssetIdMap := range newBrokerMap {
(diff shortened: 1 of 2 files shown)


apps/lib/publisher_agent/runner.go
```diff
@@ -169,7 +170,7 @@ func (r *PublisherAgentRunner[T]) RunOutgoingConnection(url BrokerPublishUrl, as
 				r.outgoingConnectionsLock.Unlock()
 			},
 		)
-		outgoingWebsocketConn := NewOutgoingWebsocketConnection[T](websocketConn, assetIds, r.logger)
```
harryrackmil, line 172: previously every time we reconnected to the stork aggregator websocket we were using local variable `assetIds` that was passed in at startup time, not the stateful `assetIds` variable on the `outgoingWebsocketConn` object.

Now we initialize the `outgoingWebsocketConn` using the local `assetIds` variable only once and on reconnect we just update the websocket connection on the same object instance.

This means that on reconnect we'll respect the latest whitelist state.

apps/lib/publisher_agent/websocket.go
```diff
@@ -237,7 +240,7 @@ func (owc *OutgoingWebsocketConnection[T]) Writer() {
 			}
 			owc.assetIdsLock.RUnlock()
 
-			if len(signedPriceUpdateBatch) > 0 {
+			if len(filteredPriceUpdates) > 0 {
```
harryrackmil, line 243: total typo - we're returning the `filteredPriceUpdates` so we should check the length of `filteredPriceUpdates`

ACK101101 (approved):
Looks like it makes sense to me.
Does the UpdateBrokerConnections function need to use the outgoingConnectionsLock? Looks like it accesses outgoingConnectionsByBroker on lines 80 and 95, and 96 is a deletion from the map

apps/lib/publisher_agent/runner.go
```diff
@@ -169,7 +170,7 @@ func (r *PublisherAgentRunner[T]) RunOutgoingConnection(url BrokerPublishUrl, as
...
+		outgoingWebsocketConn.SetWebsocketConn(websocketConn)
```
akawalsky, line 173: It seems a little weird to me that an outgoingWebsocketConn can exist without a websocket conn, and Im a bit concerned there is an edge case where the conn is read from before it exists. I'm not seeing anything immediately in the code, but does that make sense?

apps/lib/publisher_agent/runner.go
```diff
@@ -76,26 +76,30 @@ func (r *PublisherAgentRunner[T]) UpdateBrokerConnections() {
 	}
 
 	// add or update desired connections
+	r.outgoingConnectionsLock.RLock()
```
harryrackmil, line 79: add locking since we have multiple threads potentially updating this map

akawalsky (approved):

ACK101101 (approved):

PR #151, opened 2025-07-02
Use latest tag with public apps in docker

HenryMBaldwin:


akawalsky (dismissed):

HenryMBaldwin:
> Maybe dev builds should just tag as commit short instead?

This work? I kind of like having a moving dev tag as well, but I changed the main tag to be git short instead of version.txt contents

akawalsky (approved):

PR #152, opened 2025-07-07
more precise time / timestamp names

ACK101101:


ACK101101 (commented):
variables of type time.Time are named Time instead of Timestamp
variables of type int64 or str that represent Unix times are named with their correct unit

PR #156, opened 2025-07-23
Add Fuel Pusher

harryrackmil:
# Change
Support pushing Stork prices to Fuel contracts. We use Fuel's Rust SDK for contract bindings and call to the contract functions from Go.

# Testing
- [x] pusher updates the fuel proxy contract as expected


apps/lib/chain_pusher/fuel.go
```diff
@@ -0,0 +1,299 @@
...
+	"fmt"
+	"math/big"
+	"strconv"
+	"strings"
+	"unsafe"
+
+	"github.com/rs/zerolog"
+)
+
+const GasAssetId = "0xf8f8b6283d7fa5b672b530cbb84fcccb4ff8dc40f8176ef4544ddb1f1952ad07" // Fuel ETH
```
harryrackmil, line 24: Using ETH to pay gas on Fuel - not sure if that's the desired behavior.

[These docs](https://docs.fuel.network/docs/intro/glossary/#base-asset) suggest that ETH is the "base asset" and [these docs](https://github.com/FuelLabs/verified-assets/blob/main/ASSETS.md#fuel-testnet) list this as the asset id for ETH

apps/lib/chain_pusher/fuel.go
```diff
@@ -0,0 +1,299 @@
...
+		logger: logger,
+		client: client,
+	}, nil
+}
+
+func (fci *FuelContractInteractor) ListenContractEvents(
+	ctx context.Context,
+	ch chan map[InternalEncodedAssetId]InternalTemporalNumericValue,
+) {
+	fci.logger.Warn().Msg("Fuel does not currently support listening to events via websocket, falling back to polling")
```
harryrackmil, line 91: don't believe websockets are supported on fuel

apps/lib/chain_pusher/fuel_ffi/proxy_abi.json
```diff
@@ -0,0 +1,354 @@
...
+      ]
+    },
+    {
+      "inputs": [
+        {
+          "name": "update_data",
+          "concreteTypeId": "253572200a9a05ec2068dd7b1ca0b715698384a2f2031a891048be42b845b24b"
+        }
+      ],
+      "name": "update_temporal_numeric_values_v1",
```
harryrackmil, line 334: note that the proxy abi includes all the functions on the implementation

akawalsky (commented):
It all seems reasonable, but will defer to @HenryMBaldwin on Rust code. My main question is around the project organization as to whether it makes sense to have specific rust packages or to combine all pusher rust bindings to a single package.

PR #159, opened 2025-07-25
Fuel audit changes

HenryMBaldwin:
## H-01
Addressed in code and tested in 942365a85464b3f0556d1fe31f66dbd9c22d1834

## M-01
Acknowledged.

Most pushers will be running Stork's [open source chain pusher](https://github.com/Stork-Oracle/stork-external/blob/main/apps/docs/chain_pusher.md), which should mitigate this edge case. Stork would rather not introduce two way payment capabilities at this time. Stork reserves the capability to address this further in a future upgrade.

## L-01
Partially addressed in code and tested in 4a3b1fe744013b52b6be9d1b5db449287d00ffe6 (tests expanded in 05be0712e55d335cbe3d94093103e25e7438e02f).

Two step ownership transfer has been implemented, however Stork would rather not introduce ownership revocation at this time. Stork reserves the capability to implement ownership revocation in a future upgrade.

## L-02
Addressed in code in 91d20ef03d90c5f65d194bdf2ace4814d1ac769c.

Though ownership revocation is not currently possible, `revert(0)` has been added to the `revoked` arm of the `match` on the owner state as a best practice. Stork reserves the capability to revise this arm at such a time if/when ownership revocation is implemented.

## L-03
Partially address in code in a9f7ca6ce4c0da3f2c16695a09637df0f52c5af6

Recommendation 1 (signature malleability protection) has been implemented. Though signature malleability does **not** cause a vulnerability in this contract, EIP-2 compliance has been implemented in this regard as a best practice. 

Recommendation 2 (nonce in the stork signed messages) has not been implemented as the messages already have a strictly increasing nonce in the form of a timestamp which is checked in code.

## I-01
Acknowledged.

Initialization will be triggered manually immediately after deployment. As this is a one time action, the potential for griefing is minimal, especially as the deployer public key will not be publicized ahead of deployment.

### DIFF
--- a/contracts/fuel/contracts/stork/src/main.sw
+++ b/contracts/fuel/contracts/stork/src/main.sw
@@ -267,6 +267,10 @@ impl Stork for Contract {
         }
 
         let required_fee = get_total_fee(num_updates);
+        if (std::call_frames::msg_asset_id() != AssetId::base()) {
+            log(StorkError::InsufficientFee);
+            revert(0)
+        }
         if (std::context::msg_amount() < required_fee) {
             log(StorkError::InsufficientFee);
             revert(0);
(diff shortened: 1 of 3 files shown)


HenryMBaldwin:
Could potentially remove the cargo fmt from this particular pr, as it messes with the test diff maybe too much

PR #160, opened 2025-07-31
Update fuel contract to use newer features and QOL changes

HenryMBaldwin:
No significant behavioral changes have been made in this update. Arguably the only real behavioral change is the EIP-2 compliance in the `try_get_rsv_signature_from_parts` function in `verify.sw`. This **does not** remove the EIP-2 compliance, but simply cause the failed signature extraction to bubble up from the function in the form of returning `None` rather than erroring in place via `require`.
 
## New Features
- Better errors using decorators and `panic`
- Use forc.pub versions of packages
- Upgrade dependencies and toolchain versions to latest

## QOL
- Move abi def to sdk (easier integration)
- add forc-toolchain.toml to publicize toolchain versions used in this project
- Change handling of some cli commands


contracts/fuel/cli/fuels.config.ts
```diff
@@ -1,22 +1,11 @@
 import { createConfig, FuelsConfig, DeployedData, Provider, Address, Wallet, Src14OwnedProxy } from 'fuels';
 import fs from 'fs';
 
-const PRIVATE_KEY: string | undefined = process.env.PRIVATE_KEY;
-const PROVIDER_URL: string | undefined = process.env.PROVIDER_URL;
-
-if (!PRIVATE_KEY) {
```
akawalsky, line 7: Why remove these type protections?

contracts/fuel/contracts/stork/src/verify.sw
```diff
@@ -92,11 +92,9 @@ fn get_eth_signed_message_hash(msg_hash: b256) -> Message {
 fn try_get_rsv_signature_from_parts(r: b256, s: b256, v: u8) -> Option<Secp256k1> {
 
     // EIP-2 signature malleability compliance (https://eips.ethereum.org/EIPS/eip-2)
-    require( 
-        s <= 
-b256::from(0x7FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF5D576E7357A4501DDFE92F46681B20A0), 
-        "InvalidSignature" 
-    );
+    if (s > b256::from(0x7FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF5D576E7357A4501DDFE92F46681B20A0)) {
```
akawalsky, line 95: Is this logic flip intentional?

akawalsky (commented):
Just a couple comments, looks good otherwise

contracts/fuel/cli/fuels.config.ts
HenryMBaldwin: Only the deploy command actually needs these, it was annoying to have to put them in for build or dev. Also, turns out the deploy command will already complain if it doesn't like these values, so my errors were redundant.

PR #172, opened 2025-08-11
Add link audit to ci

HenryMBaldwin:


mk/lint.mk
```diff
@@ -0,0 +1,6 @@
+.PHONY: lint-links
```
akawalsky, line 1: is it necessary to make this a separate file?

HenryMBaldwin: I'm planning to put other make lint commands in here as I work through the cleanup, which is why I did this.

PR #178, opened 2025-08-14
Aaron/sto 1042 fix airstream panic with restarts

akawalsky:


apps/lib/chain_pusher/evm.go
```diff
@@ -103,64 +103,63 @@ func NewEvmContractInteractor(
...
+		err := eci.listenLoop(ctx, sub, eventCh, ch)
 		if ctx.Err() != nil {
 			return
 		}
 
-		sci.logger.Warn().Err(err).Msg("Error while watching contract events")
+		eci.logger.Warn().Err(err).Msg("Error while watching contract events")
 		if sub != nil {
 			sub.Unsubscribe()
-			close(eventCh)
```
akawalsky, line 140: Everything else in here is a naming update, this is the key changed line. We don't double close the channel now

ACK101101 (approved):

harryrackmil (approved):

apps/lib/chain_pusher/evm.go
akawalsky: True - I think we can probably get by with this for now. Moving forward we will be alerted of pusher failures.

PR #186, opened 2025-08-21
Go linting configuration

HenryMBaldwin:
golangci-lint config based on https://gist.github.com/maratori/47a4d00457a92aa426dbd48a18776322 with a few small modifications. I will not be respecting the lint in this pr, and that will be part of a separate go refactor pr, where I may also make more modifications to the lint config as I address the lint. Currently, golangci-lint finds 772 issues with this configuration.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -77,3 +77,27 @@ jobs:
               cargo clippy --all-targets --all-features --message-format json -- -D warnings 2>&1 | \
                 clippy-reviewdog-filter | \
                 reviewdog -f=checkstyle -name="lint-rust" -reporter=github-pr-check
+
+    lint-go:
+        runs-on: ubuntu-latest
+
+        steps:
+          - name: Checkout code
+            uses: actions/checkout@v4
+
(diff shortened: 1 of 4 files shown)


akawalsky:
Can you enumerate how this differs in any measurable way from internal lint rules? It would probably be worth consolidating eventually either way.

HenryMBaldwin:
> Can you enumerate how this differs in any measurable way from internal lint rules? It would probably be worth consolidating eventually either way.

- removed the depgaurd settings which were specific to packages used in internal. This will likely be modified when I respect the lint. 
- changed internal specific settings (which didn't really apply to external) for exhaustruct to a list of reasonable defaults for std and public libs. Also may change when respecting the lint. 
- Added a set of short var names (like err, ctx, etc..) that should be allowed for varnamelen. 
- Magic numbers are now allowed based on function patterns rather than based on the number in mnd.
- Stricter govet settings (this is where the majority of the gain in issue number is). Go vet the official go static analysis tool and is mostly linting for correctness.

akawalsky (approved):

PR #191, opened 2025-09-04
Cleanup pusher code

HenryMBaldwin:
I have manually confirmed all the pushers still work


akawalsky (approved):
We should integration test this before fully merging

PR #204, opened 2025-09-16
Self-Serve Stork

ACK101101:


chains/evm/contracts/first_party_stork/contracts/FirstPartyStork.sol
ACK101101, line 1: am not sure which functions should be here besides updateTemporalNumericValues

chains/evm/contracts/first_party_stork/contracts/UpgradeableFirstPartyStork.t.sol
ACK101101, line 1: this is still a work in progress

chains/evm/contracts/first_party_stork/ignition/modules/FirstPartyStork.ts
ACK101101, line 1: this is used for deploying the contracts with `npx hardhat ignition deploy ignition/modules/FirstPartyStork.ts --network localhost`

chains/evm/contracts/first_party_stork/scripts/local_register_publisher.ts
ACK101101, line 1: this is called in the docker entrypoint script - it registers the user's PUBLISHER_EVM_PUBLIC_KEY as a valid publisher for the local contract

chains/evm/contracts/first_party_stork/README.md
ACK101101, line 1: this is a placeholder

chains/evm/contracts/first_party_stork/test_viem.js
ACK101101, line 1: this is temporary

docker/scripts/first-party-evm-docker-entrypoint.sh
```diff
@@ -0,0 +1,20 @@
...
+rm -f /tmp/registered.log
+npx hardhat node &
+
+npx hardhat compile
+
+# Wait for hardhat node to initialize and then deploy contracts
+npx wait-on http://127.0.0.1:8545 && npx hardhat ignition deploy ignition/modules/FirstPartyStork.ts --network hardhatLocal --reset
+
+# Run the TypeScript registration script and log success
+npx ts-node scripts/local_register_publisher.ts && echo "REGISTRATION_COMPLETE" >> /tmp/registered.log
```
ACK101101, line 16: this is where the register script is run, it writes to a tmp file that the docker-compose healthcheck checks for. this prevents the startup race between the pusher and the contract.
the temp file is cleared out before each run on line 7

scripts/animate.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-STARTUP_ANIMATION_PATH="apps/lib/generate/frames"
+STARTUP_ANIMATION_PATH="utils/generate/pkg/frames"
```
ACK101101, line 3: this was broken

shared/signer/model.go
ACK101101, line 1: moved to shared

shared/errors.go
ACK101101, line 1: maybe overkill but i could see this being helpful in the future

Makefile
ACK101101, line 1: fixed these commands
`| cut -d: -f1` works if there are multiple GOPATH's

apps/first_party_pusher/docker-compose.yml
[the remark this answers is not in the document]
HenryMBaldwin: Could this be a profile in the root docker compose/ go into docker/compose/ or something?

apps/first_party_pusher/README.md
```diff
@@ -0,0 +1,147 @@
...
+The application consists of several key components:
+
+1. **WebSocket Server**: Handles incoming connections and price update messages
+2. **Value Processor**: Processes incoming price updates and determines when to trigger pushes
+3. **Contract Interactor**: Manages blockchain interactions and transaction submission
+4. **Asset State Manager**: Tracks the current state of each configured asset
+
+## Development
+
+### Adding New Chain Support
```
HenryMBaldwin, line 122: The "Adding New Chain Support" section seems both wrong and unnecessary.

chains/evm/sdks/first_party_stork_evm_sdk/README.md
```diff
@@ -0,0 +1,190 @@
+# FirstParty Stork EVM SDK
+
+This is a Solidity SDK to build EVM contracts that consume FirstParty Stork price feeds. This package is maintained by [Stork Labs](https://stork.network).
+
+It is available on [npm](https://www.npmjs.com/package/@storknetwork/first-party-stork-evm-sdk).
+
+## Overview
+
+The FirstParty Stork EVM SDK allows developers to interact with FirstParty Stork oracle contracts, which enable publishers to directly submit price updates on-chain. Unlike the standard Stork oracle that uses centralized aggregation, FirstParty Stork allows individual publishers to submit their own price data with configurable fees and historical data storage.
```
HenryMBaldwin, line 9: Though this is correct, it might be nice to avoid publicly calling ourselves centralized - noted that this is likely a placeholder still just shouting it out.

chains/evm/contracts/first_party_stork/foundry.toml
```diff
@@ -0,0 +1,7 @@
+[profile.default]
```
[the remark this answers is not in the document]
akawalsky: Agreed here

akawalsky (commented):
A few initial comments. Looking good, let me know when its ready for re-review after adding more tests and updating the readmes

apps/publisher_agent/pkg/runner.go
```diff
@@ -72,40 +85,56 @@ func (r *PublisherAgentRunner[T]) UpdateBrokerConnections() {
 	newBrokerMap, err := r.registryClient.GetBrokersForPublisher(publicKey)
 	if err != nil {
 		r.logger.Error().Err(err).Msg("failed to get broker connections from Stork Registry")
-		return
+	}
+
+	// merge seeded brokers with registry brokers
```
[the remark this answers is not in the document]
harryrackmil: have we tested the case where we have both a registry for a url and seeded brokers for that url? Logic seems solid but worth a unit test

apps/publisher_agent/pkg/registry.go
```diff
@@ -24,18 +25,20 @@ func NewRegistryClient(baseUrl string, storkAuthSigner signer.StorkAuthSigner, l
 }
 
 func (c *RegistryClient) GetBrokersForPublisher(
-	publisherKey signer.PublisherKey,
-) (map[BrokerPublishUrl]map[AssetId]struct{}, error) {
+	publisherKey shared.PublisherKey,
+) (map[BrokerPublishUrl]map[shared.AssetID]struct{}, error) {
+	brokerMap := make(map[BrokerPublishUrl]map[shared.AssetID]struct{})
```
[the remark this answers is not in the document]
harryrackmil: hmm I kinda prefer the current code? `return nil, error` seems pretty idiomatic in go so I'd kinda prefer to return nil rather than an empty map in case a user checked the return value for some reason. Eyeballing the code it's not obvious from `return brokerMap, fmt.Errorf("not authorized to query Stork Registry - check your configured StorkAuth")` that we're not returning a meaningful map.

Note that there are some cases where it might make sense to return a real (non-empty) map with an error (e.g. we had an error along the way but here's all the brokers that succeeded) so it's not always the case that the return object doesn't matter just because we're returning an error.

apps/first_party_pusher/docker-compose.yml
```diff
@@ -0,0 +1,151 @@
...
+      - "8080:8080"
+    volumes:
+      - "${ASSET_CONFIG_PATH:-./configs/local/pusher-asset-config.yaml}:/etc/pusher-asset-config.yaml:ro"
+    depends_on:
+      publisher-agent-local:
+        condition: service_started
+      evm-contract:
+        condition: service_healthy
+
+  first-party-pusher-production:
```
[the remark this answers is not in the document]
ACK101101: i guess it isnt hard for users to just run the pusher and agent by themselves if they want - ill remove this concept

chains/evm/contracts/first_party_stork/hardhat.config.ts
```diff
@@ -0,0 +1,56 @@
...
+          optimizer: {
+            enabled: true,
+            runs: 200,
+          },
+        },
+      },
+    },
+  },
+  networks: {
+    localhost: {
```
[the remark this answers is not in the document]
ACK101101: removed localhost

chains/evm/contracts/first_party_stork/hardhat.config.ts
```diff
@@ -0,0 +1,56 @@
...
+    },
+    hardhatOp: {
+      type: "edr-simulated",
+      chainType: "op",
+    },
+    sepolia: {
+      type: "http",
+      chainType: "l1",
+      url: configVariable("SEPOLIA_RPC_URL"),
+      accounts: [configVariable("SEPOLIA_PRIVATE_KEY")],
```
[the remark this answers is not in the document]
ACK101101: Gotcha, removing for now since it isn't being used and was more of a placeholder when I was first creating the hardhat config

ACK101101:
Addressed comments. It seems like the check-links test is a bit flakey. Anecdotally, `https://www.npmjs.com/package/@storknetwork/stork_pyth_adapter` seems to be the most problematic, but I have seen the test fail for a bunch of our different npm packages.

akawalsky (approved):

PR #206, opened 2025-09-18
Fix assetID length related panic in stark signer code

HenryMBaldwin:


shared/signer/signer.go
```diff
@@ -376,10 +387,19 @@ func createBufferFromBytes(buf []byte) *C.uint8_t {
 	return (*C.uint8_t)(unsafe.Pointer(&buf[0]))
 }
 
-func createBufferFromBigInt(i *big.Int) *C.uint8_t {
+// createStarkBufferFromBigIntAbs creates a 32-byte buffer from a big.Int and returns a pointer to the buffer in the form of a C uint8_t.
+// We enforce only use 251 bits of the buffer in order to fit into the finite field of the Stark curve.
```
harryrackmil, line 391: typo maybe?
> We enforce only use 251 bits

PR #212, opened 2025-09-30
Add Connect Methods to Contract Interactors

harryrackmil:
# Change
Pull the WS and HTTP RPC connection logic out of the contract interactor constructors and move it into dedicated `ConnectWs` and `ConnectHTTP` methods. We can then force the pusher to reconnect on failure so we don't get stuck with old rpc connections.

# Testing
- [x] pusher runs as expected
- [x] pusher reconnects to http endpoint on failure


.github/workflows/lint.yml
```diff
@@ -117,4 +117,5 @@ jobs:
               golangci_lint_flags: "--config=.golangci.yml ./apps/chain_pusher/..."
               reporter: github-pr-check
               filter_mode: nofilter
+              cache: false
```
harryrackmil, line 120: seems like there was some bad cache state for reviewdog? getting linting failures when this cache is left on - continues to fail after a successful run if I turn the cache back on. Hopefully we can turn the cache back on after this is merged though.

apps/chain_pusher/pkg/evm/interactor.go
```diff
@@ -166,13 +182,21 @@ func (eci *ContractInteractor) PullValues(
...
+
 	for _, encodedAssetID := range encodedAssetIDs {
-		storkStructsTemporalNumericValue, err := eci.contract.GetTemporalNumericValueUnsafeV1(nil, encodedAssetID)
+		storkStructsTemporalNumericValue, err = eci.contract.GetTemporalNumericValueUnsafeV1(nil, encodedAssetID)
 		if err != nil {
 			if strings.Contains(err.Error(), "NotFound()") {
 				eci.logger.Warn().Err(err).Str("assetID", hex.EncodeToString(encodedAssetID[:])).Msg("No value found")
 			} else {
 				eci.logger.Warn().Err(err).Str("assetID", hex.EncodeToString(encodedAssetID[:])).Msg("Failed to get latest value")
+				failedToGetLatestValueErr = err
```
harryrackmil, line 199: we only report an error if the error we received wasn't "No value found" since a different rpc won't fix that error

apps/chain_pusher/pkg/pusher/pusher.go
```diff
@@ -140,8 +154,6 @@ func (p *Pusher) poll(
 		polledVals, err := p.interactor.PullValues(encodedAssetIDs)
 		if err != nil {
 			p.logger.Error().Err(err).Msg("Failed to poll contract")
-
-			continue
```
harryrackmil, line 144: If we got errors when pulling some assets but successfully pulled others, we'll log an error but we'd still like to add the successful assets to the channel below. Get rid of the continue so we move on to the next if statement and add any successfully polled values.

PR #265, opened 2026-03-20
feat: add packed calldata update for ~24% calldata savings

TuDo1403:
## Summary
- Add `LibCodec` library that packs `TemporalNumericValueInput` into a flat `uint256[]` with 6 words per entry, packing `v` flag (1 bit), `timestampNs` (63 bits), and `quantizedValue` (192 bits) into a single word with zero wasted bits

## Test plan
- [x] Foundry fuzz tests (1000 runs each) for single/multi-entry encode/decode roundtrip
- [x] Calldata roundtrip via manual word construction
- [x] Edge cases: empty input, invalid length revert, invalid v revert
- [x] int192 extremes (min, max, negative values)
- [x] Calldata size comparison test: 2628 → 1988 bytes for 10 entries (~24% savings, ~8% calldata gas)
- [x] Hardhat compilation verified


chains/evm/contracts/stork/contracts/Stork.sol
akawalsky: Let's bump the version in this file to 1.0.6. That way our pusher will be able to easily determine which function to use.

This will require small test changes in UpgradeableStork.ts

akawalsky (changes requested):
Great work - left a few comments. Going to get a review from @harryrackmil as well.

harryrackmil (dismissed):

harryrackmil (dismissed):

akawalsky (approved):

harryrackmil (approved):

PR #271, opened 2026-04-08
Aaron/allow multiple signers

akawalsky:


chains/evm/contracts/stork/tasks/upgrade.ts
```diff
@@ -20,6 +20,7 @@ async function main() {
     {
       pollingInterval: 1000,
       timeout: 60000,
+      unsafeSkipStorageCheck: true,
```
akawalsky, line 23: This should be removed after this deploy. My understanding is that this should be a safe deploy because the state being added is append only, but will do extensive testing to be sure

harryrackmil (dismissed):

harryrackmil (approved):

PR #285, opened 2026-07-24
chore: handle broker registry failures better

harryrackmil:


apps/publisher_agent/pkg/runner.go
```diff
@@ -93,6 +93,7 @@ func (r *PublisherAgentRunner[T]) UpdateBrokerConnections() {
 	registryBrokers, err := r.registryClient.GetBrokersForPublisher(publicKey)
 	if err != nil {
 		r.logger.Error().Err(err).Msg("failed to get broker connections from Stork Registry")
+		return
```
harryrackmil, line 96: this is the real fix - if the agent fails to get a response from the registry they should log the error and immediately return, assuming that the brokers are the same as they were on last polling - currently if we continue meaning that we clear out the broker list and remove this connection.

This should be sufficient on its own.

apps/publisher_agent/pkg/runner.go
```diff
@@ -173,6 +165,34 @@ func (r *PublisherAgentRunner[T]) Run() {
...
+
+	for signedPriceUpdateBatch := range r.signedPriceBatchCh {
+		// copy the connections out so we never write to a connection's channel while holding the lock -
+		// a blocking write under the read lock deadlocks against onClose taking the write lock
+		r.outgoingConnectionsLock.RLock()
+		outgoingConnections := make([]*OutgoingWebsocketConnection[T], 0, len(r.outgoingConnectionsByBroker))
+		for _, outgoingConnection := range r.outgoingConnectionsByBroker {
+			outgoingConnections = append(outgoingConnections, outgoingConnection)
+		}
+		r.outgoingConnectionsLock.RUnlock()
```
harryrackmil, line 180: the `return` call above is sufficient to handle the case we saw previously, but there was also a potential deadlock which made the disconnect permanent overnight:
* we attempt to send updates to a channel whose connection has been severed so the `outgoingConnection.signedPriceUpdateBatchCh <- signedPriceUpdateBatch` blocks infinitely while holding the `r.outgoingConnectionsLock` RLock
* `UpdateBrokerConnections` runs, attempts to disconnect a connection by calling `.Remove()`, which waits for the `r.outgoingConnectionsLock` RLock to be released so it can acquire a write lock, and hangs infinitely so UpdateBrokerConnections completes and therefore never runs again

apps/publisher_agent/pkg/runner_test.go
```diff
@@ -127,6 +127,71 @@ func TestRemoveBrokerConnections(t *testing.T) {
 
 }
 
+// A broker whose writer has stopped draining its queue must not wedge the fan-out.
+// Before the fix the fan-out blocked on the full queue while holding the read lock, so onClose could
+// never take the write lock - deadlocking removal, the broker connection updater and the signing pipeline.
+func TestFanOutDoesNotBlockRemovalWhenBrokerQueueIsFull(t *testing.T) {
```
harryrackmil, line 133: test that fanout isn't blocked infinitely

apps/publisher_agent/pkg/runner.go
harryrackmil: actual fix to avoid deadlock is:
1. only hold `RLock` long enough to copy the references to the connection then release - worst case we send to a very slightly old connection for one more iteration before disconnecting
2. add a default case so we don't hang infinitely on closed/full channels

akawalsky (approved):

